PR #12, opened 2020-02-14
Reset timeout when we get a message and cleanup connect internal code

masad-frost:
Why
===
We started sending the status of the container startup from the backend when we start a connection. After discussing with the infra team (aka mason), it made sense to reset the connection timeout as long as we're receiving something from the backend, which signifies that the connection will eventually work.

While I was implementing the above, I realized how shit the code is for `tryConnect` which is the internal connect function, it had so much indirection and wasn't the most enjoyable to read, so I refactored it. I guess I could've implemented the timeout reset feature on a separate branch but the code was too smelly :hankey: 

What changed
============
First thing I did I got rid of `deferredReady` from the client. `deferredReady` was used all over the client code because we would resolve/reject the connection promise from multiple places (indirection).

Instead, now we localized all the code for resolving the promise within the `tryConnect` function itself. We're still dealing with event emitters (the websocket and channels) so it feels like there's some indirection but it's a bit more natural now since everything is colocated in the same function.

I added code comments so you can follow along the `tryConnect` function. I'll add more info this PR as well.

Test plan
=========
Try all the possible paths of failure, they all should work.

We really need to write unit tests for this as it's easily testable!

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -3,7 +3,6 @@
 import { EventEmitter } from 'events';
 import { api } from '@replit/protocol';
 import { Channel } from './channel';
-import { createDeferred, Deferred } from './deferred';
 import { EIOCompat } from './EIOCompat';
 
 enum ConnectionState {
@@ -85,8 +84,6 @@ const getWebSocketClass = (options: ConnectOptions) => {
 };
 
 export class Client extends EventEmitter {
-  public containerState: api.ContainerState.State | null;
-
   public connectionState: ConnectionState;
 
   private token: string | null;
@@ -96,8 +93,6 @@ export class Client extends EventEmitter {
   private channels: {
     [id: number]: Channel;
   };
-
-  private deferredReady: Deferred<void> | null;
 
   private debug: DebugFunc;
 
@@ -116,8 +111,6 @@ export class Client extends EventEmitter {
     this.channels = {
       0: new Channel(),
     };
-    this.deferredReady = null;
-    this.containerState = null;
     this.token = null;
     this.connectionState = ConnectionState.DISCONNECTED;
     this.debug = debug;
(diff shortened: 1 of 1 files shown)


src/client.ts
```diff
@@ -568,3 +611,30 @@ export class Client extends EventEmitter {
     ping();
   };
 }
+
+/**
+ * Emitted when there's an error while the channel is opening
+ * @asMemberOf Channel
+ * @event
+ */
+declare function close(c: { closeEvent?: CloseEvent; expected: boolean }): void;
```
masad-frost: Just typing the `close` event

src/client.ts
```diff
@@ -420,61 +386,38 @@ export class Client extends EventEmitter {
   };
 
   private onClose = ({ closeEvent, expected }: { closeEvent?: CloseEvent; expected: boolean }) => {
```
masad-frost: This kind of does the same thing except that it now is never called directly by the client unless we've had a successful connection.

The websocket cleanup is split up into a separate method so that we can call it when the `connect` function is rejected

src/client.ts
```diff
@@ -500,43 +443,164 @@ export class Client extends EventEmitter {
...
+    /**
+     * The user might call `close` before we even connect
+     * we wanna make sure we reject the promise if that happens
+     * so we monkey patch our own `close` function ;)
+     */
+    const originalClose = this.close;
+    this.close = () => {
+      this.debug({ type: 'breadcrumb', message: 'user close' });
+      onFailed(new Error('You called `Client.close` before you connected'));
+    };
```
masad-frost: Feels like we can do better in this case, but I'm not sure what.

src/client.ts
```diff
@@ -500,43 +443,164 @@ export class Client extends EventEmitter {
...
+     */
+    let onSuccess: () => void;
+    /**
+     * Failure can happen due to a number of reasons
+     * 1- Abrupt socket closure
+     * 2- Timedout connection request
+     * 3- ContainerState.SLEEP command
+     * 4- Use calling `close` before we connect
+     */
+    let onFailed: (err: Error) => void;
```
masad-frost: These methods are intitialized at the end of the function, they are referenced in the code before they are used (see `ws.onclose` for example), but since all the ways they can be called is async that shouldn't be a problem! LMK if I'm wrong

Issue #18, opened 2020-05-14
[RFC] Don't throw when channel closes during Channel::request


masad-frost:
Right now when you send a request and the channel closes; whether it is due to client closing or calling `channel.close` we throw an error for that promise. I think it would be cleaner if we just returned something instead.

Which means if you want to handle that instance you need to put `channe.request` in a try catch, catch the error 

This is common in our codebase, it's ugly and tedious!
```ts
let res
try {
  res = channel.request({ ... })
} catch (e) {
  if (/* assert if error is channel closed */) {
    handleChannelClosed();
    return;
  }

  throw e;
}
```
Not sure if this makes sense but I'm thinking this is better than throwing.
```ts
enum ChannelCloseReason {
  ClientClosed, // this means you called `Client::close`
  ClientDisconnect, // this means the websocket died unexpectedly
  ChannelClosed, // this means you called `Channel::close`
}

type RequestResult =
  | api.Command
  | { channelClosed: ChannelCloseReason; closeStatus: null }
  | { channelClosed: ChannelCloseReason.ChannelClosed; closeStatus: api.CloseChannelRes.Status };

interface Channel {
  request(cmd: api.Command): Promise<RequestResult>;
}
```

masad-frost:
So you end up with this
```ts
const res = channel.request({ ... });

if (res.channelClosed) {
  handleChannelClosed();

  return;
}
```

You can also assert the closing reason and react accordingly.

jeremypress:
I like this 👍. Is close status relevant in this case? I think we already have everything we need from the close reason.

moudy:
i like it

masad-frost:
Yeah close status is not relevant I guess. 

I went with https://github.com/replit/crosis/pull/19, it's close to this API but a little simpler. Especially because here `RequestResult` doesn't have a discriminate union!

PR #19, opened 2020-05-14
Rework Channel & Client closing signals

masad-frost:
Basically the idea is that the current channel and client closing handling sucks. More detail here https://github.com/replit/crosis/issues/18

Client:
`client.on('close'` now emits a `CloseResult` with a close reason. Diff: https://github.com/replit/crosis/commit/367ed3cbd67d08e4a7176acd5ea85fd76b7de291

Channel Close:
`channel.on('close'` now emits a `ChannelCloseReason`
Similarly `channel.close()` now resolves with `ChannelCloseReason`

Channel Request:
And the biggest change and the real impetus for this PR is that `channel.request` now doesn't throw an error when the channel closes during a request. Instead instead of resolving an `api.Command` we resolve an extended format of `api.Command` that has an optional `channelClosed` property. See `type RequestResult`

Test plan
=========
**Client:**
1. Connect client
2. attach `close` listener to client
3. call `client.close`
4. listener gets the expected close result

Repeat above, but step 3: Close websocket abruptly

**Channel Close:**
1. Connect client
2. Open channel
3. Wait for channel to open
4. Attach `close` listener on channel
4. Do steps for client closing above
5. See expected result

Repeat above, but step 4: call `channel.close`

Channel Request:
Basically do the same as above, but before closing the channel we call `channel.request`. We expect the request result to have a truthy `channelClosed` result and containers the right reasons.

### DIFF
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -3,9 +3,7 @@ import { api } from '@replit/protocol';
 import { ChannelCloseReason } from './closeReasons';
 
 type RequestResult = api.Command & {
-  channelClosed?: {
-    reason: ChannelCloseReason;
-  };
+  channelClosed?: ChannelCloseReason;
 };
 
 export class Channel extends EventEmitter {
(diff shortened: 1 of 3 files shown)


masad-frost:
Gonna merge and ship this under v5.0.0

PR #20, opened 2020-05-25
reconnect flag and open channel API

moudy:
Why
===

Currently, managing channels is tedious because the client can only be connected to once. There are a few different cases a user needs to handle when requesting a channel:

- initial setup
- cleanup after disconnecting
- setup again after reconnecting
- initial connection error

The goal of these changes is to provide an API that makes the above cases easy to handle and reason about. 

What changed
============

I modified the API to make it easier to handle the reality of network flakiness. The tests outline the intended behavior but here's an overview of the API.

```js
const client = new Client()

client.connect({
  // This will be called for every connect attempt and reconnect attempt
  token: () => Promise.resolve(REPL_TOKEN),
  reconnect: true,
}, ({ channel, error }) => {
  // Called every time cleint connects
  // Either `channel` or `error` will be null
  return (closeResult) => {
    // Called every time cleint disconnectes
    // Not called if outer function has `error`
  }
})

const close = client.openChannel({ service: 'shell' }, ({ channel, error }) => {
  // Called every time the channel opens
  // Either `channel` or `error` will be null
  return (channelCloseReason) => {
     // Called every time the channel closes
  }
})
```

Test plan
=========

Tests. I also copied this [code into our app and tested a service](https://github.com/replit/repl-it-web/compare/me-new-crosis-shell-service).


src/client.ts
```diff
@@ -128,19 +128,9 @@ export class Client extends EventEmitter {
    * Connects to the server and primes the client to start sending data
    * @returns it returns a promise that is resolved when the server is ready (sends cotainer state)
    */
-  public connect = async (options: ConnectOptions): Promise<void> => {
+  public connect = (options: ConnectOptions) => {
```
masad-frost, line 131: I think the logic in here can be simplified a little, the complexity most came from trying to make it fit into a promise API. I can do a run through after this is merged too.

src/client.ts
```diff
@@ -163,34 +153,206 @@ export class Client extends EventEmitter {
...
+        channel.handleError(err);
+      });
+
+      this.connectionState = ConnectionState.DISCONNECTED;
+      this.cleanupSocket();
+
+      this.emit('error', err);
+
+      this.debug({ type: 'breadcrumb', message: 'connect failed', data: err.message });
+    };
```
masad-frost, line 346: Hmm, should failure try to reconnect? Maybe a configurable number of times with incremental backoff?

src/__tests__/index.test.ts
```diff
@@ -0,0 +1,87 @@
...
+      // @ts-ignore: trigger unintentional disconnect
+      client.ws?.onclose();
+      disconnectTriggerd = true;
+    } else {
+      client.close();
+    }
+  });
+
+  channel.onClose((reason) => {
+    if ('clientCloseReason' in reason && reason.clientCloseReason === ClientCloseReason.Disconnected) {
```
masad-frost, line 68: You shouldn't need `in` here. I intentionally added `initiator` so that we can do pattern matching on the algebraic data type.

src/client.ts
```diff
@@ -163,34 +153,206 @@ export class Client extends EventEmitter {
...
+    /**
+     * We call this as a cleanup method after we settle the connection
+     */
+    const onFinally = () => {
+      cancelTimeout();
+      this.close = originalClose;
+      chan0.off('command', onCommand);
+    };
+
+    onSuccess = () => {
```
masad-frost, line 303: We should probably add a `connected` even here or something now that we don't use promises.

src/client.ts
```diff
@@ -387,62 +596,66 @@ export class Client extends EventEmitter {
...
-        clientCloseReason: closeResult.closeReason,
-      });
+        if (!reconnect) {
+          channel.removeAllListeners();
+        }
+      }
     });
 
     if (this.connectionState !== ConnectionState.DISCONNECTED) {
       this.emit('close', closeResult);
```
masad-frost, line 640: I'm not sure if `close` is the best event if we're `reconnecting` I think a distinction is warranted.

src/channel.ts
```diff
@@ -14,56 +28,51 @@ export class Channel extends EventEmitter {
...
 
-  public closed: boolean;
+  public options?: ChannelOptions;
 
-  // private
-  private sendQueue: Array<api.ICommand>;
+  public closed: boolean;
 
-  private sendToClient: (cmd: api.Command) => void;
+  private sendToClient: ((cmd: api.Command) => void) | null;
```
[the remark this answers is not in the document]
masad-frost: Oh nevermind, we still need to slap the channel `id` before we send to client.

moudy:
> I think the initial API you proposed in the container service makes it harder to do wrong

@masad-frost how should errors be handled when a channel fails to open with that API?

src/channel.ts
```diff
@@ -86,60 +93,82 @@ export class Channel extends EventEmitter {
...
-  public onOpen = (
-    id: number,
-    state: api.OpenChannelRes.State.CREATED | api.OpenChannelRes.State.ATTACHED,
-    send: (cmd: api.Command) => void,
-  ) => {
-    this.id = id;
-    this.sendToClient = send;
-    this.state = state;
-    this.isOpen = true;
+  public send = (cmdJson: api.ICommand) => {
```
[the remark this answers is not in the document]
moudy: yeah meant to make private. needed a way to `send` to `chan0` from the client but that could be `_send` and not exposed to user in types 🤷‍♂️

masad-frost:
Good question. I'm not 100% sure. We can have a 3rd argument for errors. Alternatively, the callback can pass `{ error: null, channel: Channel } | { error: Error, channel: null }`

moudy:
alright, i'll give it a shot. i feel like we both 180ed on this API

moudy:
@masad-frost updated API. need to work out a couple of bugs but before i do, how does this look?

src/channel.ts
```diff
@@ -14,56 +28,51 @@ export class Channel extends EventEmitter {
...
 
-    this.state = null;
+    this.options = options;
     this.id = null;
+    this.sendToClient = null;
+    this.state = null;
     this.isOpen = false;
     this.closed = false;
-    this.sendQueue = [];
-    this.sendToClient = this.enqueueSend;
```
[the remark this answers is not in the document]
moudy: The queue was for buffering messages before the channel is open. The new api only gives you a channel once it's opened so we don't need to queue anymore. But now there is a `pendingChannels` array that buffers channel requests (`client.openChannel`).

src/closeReasons.ts
```diff
@@ -24,3 +24,13 @@ export type ChannelCloseReason =
...
     };
+
+export type CloseResult =
+  | {
+      closeReason: ClientCloseReason.Intentional;
+    }
+  | {
+      closeReason: ClientCloseReason.Disconnected;
+      wsCloseEvent: CloseEvent;
+    };
```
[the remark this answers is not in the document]
moudy: moved

src/client.ts
```diff
@@ -387,62 +596,66 @@ export class Client extends EventEmitter {
...
-    delete this.channels[id];
-  };
-
-  private onClose = (closeResult: CloseResult) => {
+  private handleClose = (closeResult: CloseResult) => {
     this.cleanupSocket();
+    const reconnect =
+      closeResult.closeReason === ClientCloseReason.Disconnected &&
+      !!this.connectOptions &&
+      !!this.connectOptions.reconnect;
```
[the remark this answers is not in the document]
moudy: added some defaults

src/client.ts
```diff
@@ -387,62 +596,66 @@ export class Client extends EventEmitter {
...
+      });
+
+      this.connect(this.connectOptions);
+    } else {
+      this.channels = {
+        0: new Channel(),
+      };
+
+      this.removeAllListeners();
+    }
```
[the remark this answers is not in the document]
moudy: added a comment and made this code a little clearer

src/client.ts
moudy: simplified a little

src/client.ts
moudy: Add reconnect with backoff. I'll make it configurable. Seem like we currently fall back to polling after 2 attempts. Do you think this should continue to be handled by the user?

src/client.ts
moudy: add `client.onConnect` listener

src/client.ts
moudy: What do you think about adding `willReconnect` flag to `CloseReuslt` instead?

so 
```ts
type CloseResult =
  | { 
      closeReason: ClientCloseReason.Intentional;
    }
  | {
    closeReason: ClientCloseReason.Disconnected;
    wsEvent: CloseEvent | ErrorEvent;
    willReconnect: boolean;
  }
}
```

moudy:
realized that `connect` should probably behave that same as `openChannel` but for `chan0`. 

```js
client.connect(args, ({ channel, error }) => {

  return () => {
    // onClose
  }
});
```

working on that now

masad-frost:
Makes sense

masad-frost:
Looks great so far.

moudy:
Just pushed some changes to make `connect` behave like `openChannel` for channel 0. 

The last thing I'm unsure about is whether synchronously calling `client.close` inside `connect` callback should throw an error or if `client.close` should wrap `handleClose` in `setTimeout` so the the `connect` callback function can exit (and return it's cleanup function). You can see in the tests where I had to wrap the close functions in `setTimeout`.

This issue is unique to channel 0 since you can't close it and it doesn't get ` closeChanRes` response. Calling `channel.close` on a regular channel sends a command and waits for the `closeChanRes` response before closing so there's no issue there.

masad-frost:
Good foresight. Hmm maybe trying to make the API look exactly the same isn't the best idea? I honestly don't see why someone would call `close` inside the channel 0 callback. I can see myself calling `client.close` before or after my callback returns. I think it should be an error to close before we get a new cleanup function.

Also what's the value for `client.connect` returning a `close` function? I can see the value for the `openChannel`, there's no way to reference a `close` without having access to the channel. With the `connect` you always have a reference to `client`, hence `client.close`.

moudy:
>  I think it should be an error to close before we get a new cleanup function.

i'll try this. having the `connect` and `openChannel` API's match makes sense to me since connecting to the client basically means opening channel 0.

> Also what's the value for client.connect returning a close function? I can see the value for the openChannel, there's no way to reference a close without having access to the channel. With the connect you always have a reference to client, hence client.close.

agree. removed.

moudy:
This is getting close but I think I forgot to consider the "abort" case (user navigates away before ever connecting). We could have an explicit `abort()` method so the user's logic would look something like this

```js
function OnLeave()
 if(client.isConnected()) {
    cleint.close();
  } else {
    client.abort();
  }
};
```

But it feels much cleaner to just have `client.close` do the right thing in these cases:

```js
// Closing before ever connecting

const client = new Client()

client.connect(args, ({ channel, error }) => {
  // never called
  return () => {
    // never called
  }
})

// `close` called before connecting
client.close()
```

```js
// Closing inside open callback is the same as closing while connected (and how `openChannel` behaves)

const client = new Client()

client.connect(args, ({ channel, error }) => {
  // close inside open listener
  client.close()

  return () => {
    // called
  }
})
```

```js
// Client is closed while connected

const client = new Client()

client.connect(args, ({ channel, error }) => {
   // called
  return () => {
    // called
  }
})

// Sometime later after client connected...
client.close()
```

```js
// When a client is closed while reconnecting

const client = new Client()

client.connect(args, ({ channel, error }) => {
   // called once after initial connect
  return () => {
    // called once after dissconnect
  }
})

// Client connects...
// Client diconnects...
// `close` called while client is reconnecting...
client.close()
```

moudy:
BTW I tried this with our codebase locally and things seem to work pretty well. I think we just need to resolve the `close`/`abort` behavior outlined in the previous comment.

masad-frost:
I agree with all of it. I don't think we should have an explicit abort, the user will have to check for isConnect() which is an internal property of the so `close` for both cases makes sense.

I'm only conflicted on the behavior of this 
```
const client = new Client()

client.connect(args, ({ channel, error }) => {
  // close inside open listener
  client.close()

  return () => {
    // called
  }
})
```

Feels wrong that someone would write such code, also seems like we need a bunch of code to support this properly.

moudy:
Sounds good, I'll add some test for those cases.

> Feels wrong that someone would write such code, also seems like we need a bunch of code to support this properly.

I'll make it throw an error for now but it's not much code... just wrapping channel 0 `handleOpen` in a `setTimeout`.

moudy:
Made those updates. The last thing I'm a little unclear on is if an `openChanRes` command will always be received by the client before a command for that channel is received. I've had one of these tests fail and the only explanation I could think of was that the shell prompt message got to the client before `openChanRes` did. Is that possible?

src/client.ts
```diff
@@ -204,87 +382,121 @@ export class Client extends EventEmitter {
...
+        }
 
-    const chan0 = this.getChannel(0);
-    // avoid warnings on listener count
-    chan0.setMaxListeners(chan0.getMaxListeners() + 1);
-    // Not using Channel.request here because we want to
-    // resolve the response synchronously. We can receive
-    // openChanRes and a command on the requested channel
-    // in a single tick, using promises here would causes us to
-    // handle the incoming command before openChanRes, leading to errors
```
masad-frost, line 248: This is your problem @moudy

src/client.ts
```diff
@@ -204,87 +382,121 @@ export class Client extends EventEmitter {
...
+    const chan0 = this.getChannel(0);
 
-    // Random base36 int
-    const ref = Number(
-      Math.random()
-        .toString()
-        .split('.')[1],
-    ).toString(36);
+    chan0
+      .request({
```
masad-frost, line 429: Don't use request here as indicated by the other comment. You can be using promises, must use even emitters synchronously. This is when i discovered that `websocket.on('message'` can happen in the same tick, it was really weird time for me trying to debug the problem here.

Basically what usually happens is that, your main thread is blocked, but you're receiving frames on the socket, the browser buffers them for you and once you release the main thread the browser will call `websocket.emit('message', ...` n times for each message synchronously.

moudy: fun! will update

turbio:
@moudy lemme know if you can reproduce this
> The last thing I'm a little unclear on is if an openChanRes command will always be received by the client before a command for that channel is received. I've had one of these tests fail and the only explanation I could think of was that the shell prompt message got to the client before openChanRes did. Is that possible?

goval has strong guarantees around message ordering. If there's a bug here in goval it's pretty serious.

moudy:
@turbio pretty sure this was the issue https://github.com/replit/crosis/pull/20#discussion_r435607443

I haven't encountered again after fixing the client logic.

masad-frost:
@turbio we already addressed this, see comments above. It's js async problems

While we have your attention, how do you feel about this design?

masad-frost:
@moudy wanna update the PR description for posterity?

moudy:
yes, updated description

turbio:
gotcha thnx!

Talked through the design with @masad-frost, sounds pretty good to me. Reconnecting is gonna be tricky no matter what so I think forcing the lib user to handle it explicitly makes the most sense. Ordering and synchronous behavior should probably be well documented and enforced. Since we have two different constructs called channels I'm going to refer to them as logical channels and goval channels... but maybe we should rename?


style nits that aren't very important:
- I think we should drop support for the checking goval channel close status. Since creation/destruction must already be handled, it seems natural that this is the consumers responsibility.
- The multi callback pattern looks a bit odd. Seems like really we're creating a constructor/destructor with an  unusual pattern. It also imposes non obvious (maybe it's intentional and I'm understanding though?) constraints on code being synchronous.

Some of the non obvious behaviors we should make clear:
- do anonymous logical channels reconnect?
- what happens when you call close on a logical channel in your close handler? is that even allowed? Is it different if the close was intentional (calling close) or unintentional (network error).
- the close handler MUST synchronously clean up everything before returning. We should do as much as we can to prevent/warn against usage and storage of the channel after this point.
- the order of close handler and request resolutions from close should be well defined

moudy:
@turbio think I'm following... but can you clarify what "logical channels and goval channels" are? 

> I think we should drop support for the checking goval channel close status. Since creation/destruction must already be handled, it seems natural that this is the consumers responsibility.

the `isConnected` function right? yeah we can remove that

> The multi callback pattern looks a bit odd. Seems like really we're creating a constructor/destructor with an unusual pattern. It also imposes non obvious (maybe it's intentional and I'm understanding though?) constraints on code being synchronous.

Is there a more conventional constructor/destructor pattern? I like how this locates setup/cleanup logic together but agree it's not the most common pattern (although in the JS world React's `useEffect` hook makes this less foreign). `useEffect` has the same issue with code being synchronous. I'm not sure if the client has enough knowledge to know what to do if the open callback could be async. For example, if the client closes for some reason while in the middle of the async callback being executed what happens? 

> do anonymous logical channels reconnect?

Yes, if I'm correctly understanding what these are. 

> what happens when you call close on a logical channel in your close handler? is that even allowed? Is it different if the close was intentional (calling close) or unintentional (network error).

Gonna think about his and add a test case. My initial reaction is that it should throw an error. I forgot to include this in the description... the close callback for client and channel both get passed an argument with info about why they closed (intentional or disconnect).

> the close handler MUST synchronously clean up everything before returning. We should do as much as we can to prevent/warn against usage and storage of the channel after this point.

Agree, I think this API kind of encourages that since you'd have to store a reference to that channel in the outer scope. Not sure what else to do other than docs and throw some errors.

> the order of close handler and request resolutions from close should be well defined

Yeah, I'll add docs/tests around the order of these events.

turbio:
logical channel being the object you get back from crosis which manages reconnecting and can be closed. The goval channel is the thing passed to the constructor and can be destroyed at any point.

The required scope is cool, if it fits other common js patterns then it sounds like a good idea. Looks like the lib encourages proper usage which I imagine is super important. My thinking on removing `isConnected` is that it further discourages passing the goval channel around, anything we can do to keep usage correct sounds good.

a good first test of the new interface would be updating repl.it/@util/replrun to use it. It's a nice simple self contained goval client.

src/client.ts
```diff
@@ -238,53 +432,86 @@ export class Client extends EventEmitter {
...
   public close = () => {
     this.debug({ type: 'breadcrumb', message: 'user close' });
 
-    this.onClose({ closeReason: ClientCloseReason.Intentional });
+    if (!this.connectOptions) {
+      throw new Error('Must call client.connect before closing');
+    }
+
+    if (this.isOpenningChan0) {
+      throw new Error('Cannot call close while connecting');
```
masad-frost, line 506: I we can use better words here. `Cannot call closing inside connect callback`?

src/client.ts
masad-frost: Yeah makes sense

src/client.ts
```diff
@@ -238,53 +432,86 @@ export class Client extends EventEmitter {
...
    */
   public close = () => {
     this.debug({ type: 'breadcrumb', message: 'user close' });
 
-    this.onClose({ closeReason: ClientCloseReason.Intentional });
+    if (!this.connectOptions) {
+      throw new Error('Must call client.connect before closing');
+    }
+
+    if (this.isOpenningChan0) {
```
masad-frost, line 505: I we can use better words here. `Cannot call close inside connect callback`?

src/client.ts
```diff
@@ -89,17 +115,27 @@ export class Client extends EventEmitter {
...
   private channels: {
     [id: number]: Channel;
   };
 
   private debug: DebugFunc;
 
-  private didConnect: boolean;
+  private retryTimer: ReturnType<typeof setTimeout> | null;
+
+  private connectTrys: number;
```
masad-frost, line 134: `connectTries`

src/client.ts
```diff
@@ -387,62 +596,66 @@ export class Client extends EventEmitter {
...
-  private onClose = (closeResult: CloseResult) => {
+  private handleClose = (closeResult: CloseResult) => {
     this.cleanupSocket();
+    const reconnect =
+      closeResult.closeReason === ClientCloseReason.Disconnected &&
+      !!this.connectOptions &&
+      !!this.connectOptions.reconnect;
+
+    Object.values(this.channels).forEach((channel) => {
+      if (channel.isOpen) {
```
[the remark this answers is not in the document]
masad-frost: Shouldn't this be an error?

src/client.ts
```diff
@@ -204,32 +382,48 @@ export class Client extends EventEmitter {
...
+    const channel = new Channel({ chanReq, options });
+
+    if (chan0.isOpen) {
+      // We're connected, open channel
+      this.handleOpenChannel(channel);
+    } else {
+      this.pendingChannels.push(channel);
+    }
+
+    return () => channel.close();
```
masad-frost, line 398: It doesn't seem like we handle calling close on "pending" channels.

src/client.ts
```diff
@@ -238,53 +432,86 @@ export class Client extends EventEmitter {
...
+      if (this.connectionState !== ConnectionState.CONNECTED) {
+        channel.handleError(new Error('Client not connected'));
+      }
+
       if (cmd.openChanRes == null) {
         throw new Error('Expected openChanRes on command');
       }
 
-      this.handleOpenChanRes(channel, cmd.openChanRes);
+      dispose();
```
masad-frost, line 465: Should probably dispose as soon as we get a command that matches our ref.

src/client.ts
```diff
@@ -238,53 +432,86 @@ export class Client extends EventEmitter {
...
     });
 
-    const onResponse = (cmd: api.Command) => {
+    const dispose = chan0.onCommand((cmd: api.Command) => {
       if (ref !== cmd.ref) {
         return;
       }
 
+      if (this.connectionState !== ConnectionState.CONNECTED) {
+        channel.handleError(new Error('Client not connected'));
```
masad-frost, line 458: Awkward, how can this happen?

src/client.ts
```diff
@@ -39,11 +48,28 @@ type DebugLog =
 type DebugFunc = (log: DebugLog) => void;
 
 interface ConnectOptions {
-  token: string;
-  urlOptions?: UrlOptions;
-  polling?: boolean;
-  timeout?: number | null;
+  token: () => Promise<string>;
```
masad-frost, line 51: This should probably be `getToken` or `fetchToken`

src/client.ts
```diff
@@ -39,11 +48,28 @@ type DebugLog =
...
+interface ConnectArgs extends Partial<Omit<ConnectOptions, 'token'>> {
+  token: () => Promise<string>;
+}
+
+const backoffFactor = 1.7;
+const maxBackoff = 15000;
+function getNextRetryDelay(retryNumber: number) {
+  const randomMs = Math.floor(Math.random() * 500);
+  // TODO: why no Math.pow?
+  // eslint-disable-next-line no-restricted-properties
```
masad-frost, line 69: I think it wants you to use `**` we should disable the lint rule or use that.

src/client.ts
masad-frost: I'm not sure to be honest. Yeah let's just go with polling after 2 tries for now.

src/client.ts
moudy: Can you elaborate? In what case does the user get a "pending" channel?

masad-frost: Scenario:
Client is not connected
User calls `openChannel`
User gets the dispose function back
User calls the dispose function
Client connects
The channel still gets openned

src/client.ts
```diff
@@ -397,293 +628,150 @@ export class Client extends EventEmitter {
...
-        clientCloseReason: closeResult.closeReason,
-      });
+    this.connectToken = null;
+
+    if (this.retryTimer) {
+      // Client was closed while reconnecting
+      clearTimeout(this.retryTimer);
+    }
+
+    Object.values(this.channels).forEach((channel) => {
```
masad-frost, line 701: We should move these channels to `pendingChannels` and `this.channels` should be empty IMO

src/client.ts
```diff
@@ -397,293 +628,150 @@ export class Client extends EventEmitter {
...
+      }
+    }
 
-    delete this.channels[id];
+    // Open existing channels when we connect
+    Object.values(this.channels).forEach((channel) => {
+      if (channel.options) {
+        this.handleOpenChannel(channel);
+      }
+    });
```
masad-frost, line 688: And then we remove this part.

src/client.ts
```diff
@@ -238,53 +429,86 @@ export class Client extends EventEmitter {
...
-    return channel;
+      if (id == null || state == null) {
+        throw new Error('Expected state and channel id');
+      }
+
+      if (channel.id) {
+        // Remove old channel from map. It gets added back with a new id right after this block
+        // This happens when client reconnects
+        delete this.channels[channel.id];
+      }
```
masad-frost, line 482: Also we don't need this, `id`'s are not guaranteed to be the same. This logic leaks channels

src/client.ts
```diff
@@ -89,17 +114,27 @@ export class Client extends EventEmitter {
...
 
-  private token: string | null;
-
   private ws: WebSocket | null;
 
+  private connectOptions: ConnectOptions | null;
+
+  private chan0Req: ChanReqFn | null;
+
+  private pendingChannels: Array<Channel>;
```
masad-frost, line 123: We should probably even pend the channel `ChannelOptions` not the channel itself.

src/client.ts
moudy: Not sure I understand. When a channel reconnects don't we want to update the key to match the new id? Or should we not expose `getChannel`?

src/client.ts
moudy: Can we make these types of updates after publishing? Being able to use this new API is blocking a lot of workspace work.

src/client.ts
```diff
@@ -163,34 +191,181 @@ export class Client extends EventEmitter {
...
+            // Once we're READY we can stop listening to incoming commands
+            dispose();
+
+            if (this.retryTimer) {
+              clearTimeout(this.retryTimer);
+            }
+            cancelTimeout();
+
+            // To
+            this.isOpenningChan0 = true;
```
masad-frost, line 326: I personally like to monkey patch my own object in these cases, feels like less indirection and less state.

src/client.ts
```diff
@@ -163,34 +191,181 @@ export class Client extends EventEmitter {
...
+
+            chan0.handleOpen({
+              id: 0,
+              state: api.OpenChannelRes.State.CREATED,
+              send: this.send,
+            });
+
+            this.isOpenningChan0 = false;
```
masad-frost, line 334:
```suggestion
            const originalClose = this.close;
            this.close = () => throw new Error('Cannot call close inside connect callback');

            chan0.handleOpen({
              id: 0,
              state: api.OpenChannelRes.State.CREATED,
              send: this.send,
            });

            this.close = originalClose;
```

src/client.ts
masad-frost: Sorry I misunderstood the code here!

Either way, I thought we talked about getting a new channel every time we reconnect.

src/client.ts
masad-frost: Publish whenever you want

src/client.ts
moudy: With this implementation, you get the same channel object back that it has had all it's listener removed and has a new id and send function. I initially had "channelRequest" objects that stuck around between disconnects but it seemed like those channels themselves could serve the same function.

masad-frost: Feels like a trap ready to happen. I think channels shouldn't be reused. It's actually kinda scary

moudy: I don't see the issue. Can you give some examples?

turbio: If I'm understanding correctly you're reusing the goval channel? Besides being a trap (You can actually use it after it's been destroyed), it prevents strong asserts on how the channel is used. We won't be able to enforce usage which I think is super important to get this right

moudy: I'm resuing the channel object. You could always use a channel after it's destroyed if your logic doesn't account for the channel closing (regardless if it's the same object reference between connections). Can you give some examples or test cases of how this is a trap? I'm still confused.

masad-frost: To start with, you're really openning a new channel to goval and channels are stateful.

Another thing is that your callback will be the "setup" function. You will be setting up routines (async functions and timers) in there, these things need to know to stop and give up once the channel closed even if the client reconnected. Otherwise you need to do some work to know what needs to be re-setup and what doesn't, complicates things

Here's a bit of a contrived example that has to do with async

```js
async function doStuff(channel) {
  let output = '';
  channel.onCommand((cmd) => {
     if (cmd.output) output += cmd.output;
   });

  await channel.request({ someRequestWithOutput: {} })

  // channel might disconnect (and reconnect with the current implementation) while doing the request
  await doHttpRequest()

  if (channel.closed) {
    // channel might've reconnected here with our current logic
    return;
   }

  // this request relies on the previous command completing
  await channel.request({ anotherRequestThatReliesOnGovalChannelState: { } })
  await channel.request({ anotherRequestWithOutput: {} })

  // output here doesn't really match the current channel's output
  doSomethingWithChannelOutput(output);
}
```

masad-frost: > You could always use a channel after it's destroyed if your logic doesn't account for the channel closing

2 things about this. Is that if you use a closed channel, an error will be thrown, secondly, you can't really assert whether the channel is the same goval channel or not (i.e. a reconnected channel).

moudy: Alright will try to figure out soon

src/client.ts
moudy: Don't think it can. I can remove this check.

src/client.ts
moudy: This should be fixed by the `channelRequest` object being removed from `this.channelRequests` [here](https://github.com/replit/crosis/blob/6c88e68bb947665295ec2c2917c91056189670c2/src/client.ts#L407)

moudy:
@masad-frost converted to using new channel objects and addressed some more feedback related to handling pending channels. going to publish another beta

src/closeReasons.ts
```diff
@@ -18,7 +18,13 @@ export enum ClientCloseReason {
...
   | {
       initiator: 'client';
-      clientCloseReason: ClientCloseReason;
+      clientCloseReason: ClientCloseReason.Disconnected;
+      willReconnect: boolean;
+      wsEvent: CloseEvent | ErrorEvent;
+    }
+  | {
+      initiator: 'client';
+      clientCloseReason: ClientCloseReason.Intentional;
```
masad-frost, line 27:
```suggestion
      willReconnect: boolean;
```

masad-frost: Do we need this information at the channel level? My feeling is that we won't use it
```
export type ChannelCloseReason = { intiator: 'client', willReconnect: boolean} | { initiator: 'channel', willReconnect: false, closeStatus: api.CloseChannelRes.Status; }

moudy: yeah prob don't need it. also, do you think we need `closeStatus` info on channel close?

masad-frost: Probably not

masad-frost: Probably not

masad-frost:
Merging into v6 branch and we can base off of that until we get out of beta

PR #22, opened 2020-07-13
Split internal connect function from external

masad-frost:
Why
===
Having the external function and the internal reconnect function be the same makes it hard for us to reason about reconnects and stuff.

What changed
============
- Renamed public `connect` function to `open` (we can keep the public API as `connect`)
- Made a private function called `connect` that is used internally to reconnect and connect for the first time
- Client.connectOptions is not null anymore, it's always initialized, `fetchToken` is initialized with a promise that rejects
- Set `this.chan0Cb` when we disconnect without reconnecting
- Use `this.chan0Cb` to make sure the user is not calling `open` multiple times without closing

Test plan
=========
Tests pass

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "eslint-plugin-react": "^7.14.3",
     "jest": "^26.0.1",
     "parcel": "^1.12.4",
-    "prettier": "^1.18.2",
+    "prettier": "^2.0.5",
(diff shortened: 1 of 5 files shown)


masad-frost:
I also updated prettier because it wasn't picking up some typescript stuff

PR #23, opened 2020-07-14
add retry breadcrumb

moudy:
Why
===

debugging case where reconnect can happen when already connected

What changed
============

added breadcrumb

Test plan
=========

extra data should help with debugging


moudy:
We use it as a flag to send to sentry here https://github.com/replit/repl-it-web/blob/master/app/workspace/plugins/adminpanel/CrosisV6Logs.tsx#L20

PR #28, opened 2020-08-07
skip channel open option

moudy:
Why
===

I described the general problem in https://github.com/replit/crosis/pull/27. I think this change is still needed because I think it's what you would expect. But had something better in mind

Based on the user/repl connected to the container some services are relevant while others are not. And since we want to reuse the same client between connections there needs to be a way to conditionally open a channel when we connect. This could be accomplished by allowing the user to provide a `skip` function that prevents the channel from opening if it returns true. 

The client's interface is a stream of channels so `skip` can be thought of as an on/off switch.

What changed
============

Added a `skip` option to `openChannel`.

Test plan
=========

test/feedback

### DIFF
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -14,11 +14,15 @@ export type OpenChannelRes = { error: null; channel: Channel } | { error: Error;
  * This function gets called when a channel opens or there is an error opening.
  * It can return a function that can be used to cleanup an logic when the channle closes.
  * If there is an error opening the channel the cleanup function is not called. You can
- * think of it as a stream of values that terminates if/when there is an error.
+ * think of it as a stream of values that terminates if/when there is an error. A `skip`
+ * function may be provided to skip opening the channel conditionally on every reconnection.
  *
  * Example:
  *
- * const closeChannel = client.openChannel({ service: 'shell' }, ({ channel, error }) => {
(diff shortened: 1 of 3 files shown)


src/channel.ts
```diff
@@ -199,7 +200,7 @@ export class Channel extends EventEmitter {
     if (this.openChannelCbClose) {
       this.openChannelCbClose(reason);
       this.openChannelCbClose = null;
-    } else {
+    } else if (!reason.willReconnect) {
```
moudy, line 203: The test I added surfaced another bug. When the user doesn't provide a cleanup function but the client will reconnect there's no reason to call `openChannelCb` with an error. I should probably write an explicit test for this.

src/client.ts
```diff
@@ -241,7 +244,11 @@ export class Client extends EventEmitter {
   };
 
   private handleOpenChannel = (channelRequest: ChannelRequest) => {
-    const { options, openChannelCb } = channelRequest;
+    const { options, openChannelCb, skip } = channelRequest;
+
+    if (skip && skip()) {
```
masad-frost, line 249: Why do we have 2 run the same check in 2 places?

moudy: right, removed. we just need the one in `handleOpenChanne`

PR #34, opened 2020-08-18
[v5] Call close even if the channel never opened

masad-frost:
Why
===
Channels should get a signal that they're closed regardless of whether they ever opened or not. Getting a channel is optimistic and you can start sending requests on it and use it, so we need to do proper cleanup (i.e. rejecting all requests) when the client closes before the channel actually opens.

What changed
============
- Channels that are inflight get added to a set of `this.openningChannels` while their open requests are inflight.
- When the we close the connection we call the close handler on each of the unoppened channels

Test plan
=========
run this code
```
const c = new Crosis.Client();

c.connect({
  token: TOKEN_HERE,
}).then(() => {
  console.log('connected');
  const chan = c.openChannel({ name: 'exec' });
  chan.on('open', () => {
    console.log('channel opened');
  });
  chan.on('close', () => {
    console.log('chan close called');
  });

  c.ws.close();
});
```
Notice that `'chan close called'` is logged without `'channel opened'` being called

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -97,6 +97,8 @@ export class Client extends EventEmitter {
     [id: number]: Channel;
   };
 
+  private inflightChannels: Set<Channel>;
+
   private debug: DebugFunc;
 
   private didConnect: boolean;
@@ -114,6 +116,7 @@ export class Client extends EventEmitter {
     this.channels = {
       0: new Channel(),
     };
+    this.inflightChannels = new Set();
     this.token = null;
     this.connectionState = ConnectionState.DISCONNECTED;
     this.debug = debug;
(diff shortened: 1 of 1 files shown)


src/client.ts
```diff
@@ -438,6 +440,16 @@ export class Client extends EventEmitter {
...
     });
 
+    // These are channels that never opened, but we should emit the
+    // close event on them anyway to do proper cleanup
+    this.inflightChannels.forEach((chan) => {
+      chan.onClose({
+        initiator: 'client',
+        clientCloseReason: closeResult.closeReason,
+      });
+      this.inflightChannels.delete(chan);
```
masad-frost, line 450: Not sure if we should delete them here since the client can't be re-used anyway

masad-frost: We just lose some information on the dead client but it's fine

moudy:
is this is the equivalent of the v6 api being called with an error on initial open?

masad-frost:
Yeah pretty much!

src/client.ts
moudy: seems fine

PR #35, opened 2020-08-24
RFC: connection context

moudy:
Why
===

A common thing that's coming up is getting repl/user info where channels are being opened. For example, you might want to skip opening a channel based on auth state or a repl's language capabilities. 

What changed
============

Adds an optional `context` argument when opening a client that gets passed to `skip` and `openChannel` functions.

Test plan
=========

test/feedback

### DIFF
--- a/src/__tests__/index.test.ts
+++ b/src/__tests__/index.test.ts
@@ -414,3 +414,39 @@ test('closing client while opening', (done) => {
     },
   );
 });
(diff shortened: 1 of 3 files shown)


masad-frost:
Big fuck yes. I love it.

masad-frost:
@sergeichestakov thoughts?

PR #37, opened 2020-08-31
fix generic typing on context object

moudy:
Why
===

there isn't a good way to type the `open` and `openChannel` methods

What changed
============

made those methods generic

Screen Shot 2020-08-31 at 8 48 38 AM

verifying that the context passed in matches the callback's context type
Screen Shot 2020-08-31 at 8 47 36 AM

Screen Shot 2020-08-31 at 8 47 48 AM

Test plan
=========

test still pass

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -55,7 +55,7 @@ interface ConnectOptions<D = any> {
   reconnect: boolean;
   WebSocketClass?: typeof WebSocket;
   maxConnectRetries: number;
-  context?: D;
+  context: D;
 }
 
 interface ChannelRequest {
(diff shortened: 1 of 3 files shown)


src/client.ts
```diff
@@ -313,7 +314,10 @@ export class Client extends EventEmitter {
 
       if (state === api.OpenChannelRes.State.ERROR) {
         this.debug({ type: 'breadcrumb', message: 'error', data: error });
-        channel.handleError(new Error(error || 'Something went wrong'));
+        channel.handleError(
+          new Error(error || 'Something went wrong'),
+          this.connectOptions.context,
```
moudy, line 319: also fixed context not being passed to error callback

sergeichestakov: should be separate PR next time

PR #41, opened 2020-09-04
fix utf8 strings (again!)

turbio:
Why
===

They're still broken man. Finally had some time to go spelunking around the pbjs utf8 decoder. Turns out they're super clowny :clown_face: 

What changed
============

basically just remove the weird chunking code.

Test plan
=========

after lots of fuzzing and benchmarking, only prod can tell us what's what.


masad-frost (approved):
If you can document some of what's going on and why is their original read function borked

PR #43, opened 2020-09-07
clear previous retry timer when connection fails

moudy:
Why
===

so the client doesn't try to connect again when closes intentionally. 

What changed
============

clear previous retry timer when the connection fails. if the client is configured to re-connect a new retryTimer will be created just below this

Test plan
=========

test pass and shouldn't see `Expected chan0Cb` error anymore

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -595,6 +595,10 @@ export class Client {
       });
 
       onFailed = (error: Error) => {
+        if (this.retryTimer) {
+          clearTimeout(this.retryTimer);
+        }
+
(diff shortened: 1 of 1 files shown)


sergeichestakov (approved):

sergeichestakov (commented):
Actually doesn't this timeout get cleared in the call to `handleConnectError` below?

moudy:
handleConnectError is called once we give up trying to reconnect so my theory here is that what's missing is clearing in between connection attempts

PR #45, opened 2020-09-12
Fallback to polling if connecting via WS fails

sergeichestakov:
Why
===

We should be falling back to polling if attempting to connect via Websocket fails on every attempt. This will help reduce timeout errors and at least establish some sort of connection. It also matches what we do with Crosis v5 in the IDE.

What changed
============

Fallback to polling if we've failed to connect via WS `maxConnectRetries` amount of times.

Test plan
=========

- Test I added passes
- Should see errors in Sentry go down


moudy (approved):
👍

PR #47, opened 2020-09-17
Don't silently fail token requests

masad-frost:
Why
===
`fetchToken` can silently fail with no recourse path. We need to report it

What changed
============
When fetch token function fails, we now call `handleConnectError` with the error, which in turn calls the channel 0 callback with the error

Test plan
=========
Wrote a test that runs through the code path. Other tests still work.


moudy (approved):
🤕

PR #48, opened 2020-09-17
Add debug breadcrumb for every connection attempt

sergeichestakov:
Why
===

This will help us gain more insight into the total number of connection attempts, and why they fail when they do.

What changed
============

Add debug breadcrumb that gets called on every call to `connect`

Test plan
=========

We'll see this logged next time we publish


dan-stowell (approved):
Hawt! 🌋

PR #49, opened 2020-09-17
Dont use EIOCompat WS in test env

sergeichestakov:
Why
===

EIOCompat WS does not work in Jest env. Let's not use it in tests.

What changed
============

Check for `process.env.NODE_ENV` before returning EIO WS when polling

Test plan
=========

"client errors opening" test no longer fails


sergeichestakov:
cc @masad-frost

masad-frost (approved):
That's fine. LEt'ss fix the tests

PR #50, opened 2020-09-17
Add fatal function to crosis

masad-frost:
Why
===
Thrown errors outside of a synchronous call are thrown into the aether and handled by the global error handler. We need a way to probe these errors and give the client's user to react to it properly.

What changed
============
- Added fatal function that's passed to the constructor
- Call fatal and return instead of throwing in places that can't be handled in a try catch around a call on the client
- Removed debug function from constructor since we just use `setDebugFunc` anyway

Test plan
=========
I added a simple test, it works. Will dig deeper into other errors

### DIFF
--- a/src/__tests__/index.test.ts
+++ b/src/__tests__/index.test.ts
@@ -14,7 +14,8 @@ if (!REPL_TOKEN) {
 jest.setTimeout(10 * 1000);
 
 test('client connect', (done) => {
-  const client = new Client();
+  const fatal = jest.fn<void, [Error]>();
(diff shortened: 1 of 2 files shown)


sergeichestakov (changes requested):
I'm not a fan of this API nor do I think we should be making `fatal` required/an argument to the constructor. Can we use event emitters and expose an `onError` method?

masad-frost:
will follow up

sergeichestakov:
Let's not merge breaking changes without approval. Going to revert.

PR #51, opened 2020-09-17
Abort inflight token requests when closing

masad-frost:
Why
===
We can end up in a really awkward state where we call `client.close` while the client is fetching a token. When that happens, we think things are cleaned up but when the token fetch request completes we end up we throw an error and the client gets into a bad state.

What changed
============
`fetchToken` now gets an abort signal that can be used to be listened to abort or passed to a fetch function

Test plan
=========
Wrote a test for it

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -78,6 +78,8 @@ export class Client {
 
   private debug: DebugFunc;
 
+  private fatal: (e: Error) => void;
+
   private retryTimeoutId: ReturnType<typeof setTimeout> | null;
 
   private connectTries: number;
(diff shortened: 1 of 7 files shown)


src/index.ts
```diff
@@ -1,3 +1,4 @@
+import 'abortcontroller-polyfill/dist/abortcontroller-polyfill-only';
```
masad-frost, line 1: Should we ponyfill instead?

src/client.ts
```diff
@@ -757,16 +826,10 @@ export class Client {
...
-            willReconnect: false,
-          }
-        : {
-            initiator: 'client',
-            willReconnect,
-          };
+    const closeReason: ChannelCloseReason = {
+      initiator: 'client',
+      willReconnect,
+    };
```
masad-frost, line 832: unrelated change slipped through, but the logic seemed unnecessarily duplicated

src/client.ts
```diff
@@ -780,10 +843,7 @@ export class Client {
 
     this.connectionState = ConnectionState.DISCONNECTED;
 
-    if (
-      closeResult.closeReason === ClientCloseReason.Intentional ||
-      !this.connectOptions.reconnect
-    ) {
+    if (!willReconnect) {
```
masad-frost, line 846: same

masad-frost:
@sergeichestakov take a look when you can, gonna merge and keep going

PR #52, opened 2020-09-18
Slim down options surface area and fix some error handling

masad-frost:
Why
===
Options makes code harder. More concretely there was a problem with the chan0Cb being called twice when we run out of retries. We also had a problem where the client could potentially keep trying to reconnect if we call `client.close`. We are not really using those options and we wanna retry forever anyway

What changed
============
- Removed retries options, we currently set them to default. In reality we wanna retry forever until the client.close is called
- Removed polling options as it's based on maximum retries and that's removed (will revisit in the future)
- onUnrecoverableError now calls handleClose with a new close reason (Error)
- chan0Cb is not called twice when there's a an error connecting
- Calling `client.close` after the websocket connection fails doesn't keep retrying

Test plan
=========
Will follow up with a separate PR for more elaborate tests. The abort & polling tests now fails, but that's it.

For now I built dev and went to the browser to try things manually
- The client indefinitely reconnects when the handshake/ws fails
- The client eventually connects when the handshake works
- Unrecoverable errors cause the client to close and not try to reconnect

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -23,6 +23,10 @@ type CloseResult =
   | {
       closeReason: ClientCloseReason.Disconnected;
       wsEvent: CloseEvent | ErrorEvent;
+    }
+  | {
+      closeReason: ClientCloseReason.Error;
+      error: Error;
     };
 
 enum ConnectionState {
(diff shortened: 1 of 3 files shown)


turbio (approved):
neeto

PR #53, opened 2020-09-18
Fix client generic context

masad-frost:
Why
===
Right now context is passed to each `open` and `openChannel` calls. Typing like that works for most cases but you don't get any real type check guarantees. As it's unlikely for the shape of the context to change between `open` calls, we set the context on the client and let 

What changed
============
Ctx is now passed as a generic to `Client<{ username: string }>`. All calls now implicitly get the context.
Renamed generic from `D` to `Ctx`
`connectOptions` is now null until `open` is called

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -63,7 +63,7 @@ interface ChannelRequest<Ctx> {
   openChannelCb: OpenChannelCb<Ctx>;
 }
 
-export class Client<Ctx extends unknown = null> {
+export class Client<Ctx extends unknown = void> {
   public static ClientCloseReason = ClientCloseReason;
 
   public connectionState: ConnectionState;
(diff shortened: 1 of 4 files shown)


src/client.ts
```diff
@@ -115,6 +115,11 @@ export class Client {
         port: '80',
       },
       fetchToken: () => Promise.reject(new Error('You must provide a fetchToken function')),
+      // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+      // @ts-ignore context is not relevant until we call the `open` function
```
turbio, line 119: eek, this feels wrong since it's not a consequence of the environment. IMO tslint isn't really appropriate like this in new code :shrug:

turbio (changes requested):

src/client.ts
masad-frost: Can you expand on what you mean? There's no way to deal with this type of generic otherwise, the only alternative is what we had or maybe I'm missing something. What we had is an in-cohesive context type that doesn't give any static typing guarantees.

masad-frost: @sergeichestakov can I get your input here?

Happy to close the PR but this is more useful than using `any` all over the place and in-cohesive context

sergeichestakov: yeah it's not great. any way we can make `context` an optional param? probably should be since we don't require the user to pass it in

masad-frost: Made it optional. Gonna try it out later and see how it feels.

masad-frost: Isn't really usable like this, doesn't inherit the actual type from the client, assumes it could always be falsy. Gonna try another approach

masad-frost:
Updated the description

sergeichestakov (approved):

PR #54, opened 2020-09-18
Stronger asserts around opening & closing channels

masad-frost:
Why
===
There aren't any real asserts that stop us from doing bad things in the channel and with the channel

What changed
============
- Channel gets passed `onUnrecoverableError` function from the client
- Client now maintains a single reference to `onUnrecoverableError`. This allows us to pass the function to channels and make sure we trigger the right handler if the user changes their recoverable error handler function using `Client.setUnrecoverErrorHandler`
- Errors that are normally thrown now call `onUnrecoverableError` inside the channel
- Added some invariant checks that trigger `onUnrecoverableError`
- Avoid monkey patching methods on the channel upon closing in favor of inline checks for `closed`

Test plan
=========
- Open a channel
- Close socket
- Keep reference to old channel
- Calling `request`/`send`/`onCommand` triggers an unrecoverable error

Will follow up with tests for this & https://github.com/replit/crosis/pull/52

### DIFF
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -81,6 +81,13 @@ export class Channel<Ctx> {
   }
 
   public onCommand = (listener: (cmd: api.Command) => void) => {
+    if (this.closed) {
+      const e = new Error('Trying to listen to commands on a closed channel');
+      this.onUnrecoverableError(e);
+
+      throw e;
+    }
+
     this.emitter.on('command', listener);
 
     return () => this.emitter.removeListener('command', listener);
@@ -115,8 +122,7 @@ export class Channel<Ctx> {
 
     // Send close command to chan0
     this.sendToClient(cmd);
-
-    this.closeInternal();
+    this.closed = true;
   };
 
   /**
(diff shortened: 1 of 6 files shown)


src/channel.ts
```diff
@@ -89,7 +101,9 @@ export class Channel<Ctx> {
    */
   public close = (action: api.CloseChannel.Action = api.CloseChannel.Action.TRY_CLOSE) => {
     if (this.closed === true) {
-      throw new Error('Channel already closed');
+      this.onUnrecoverableError(new Error('Channel already closed'));
+
+      return;
```
turbio, line 106: should you throw here too?

src/channel.ts
```diff
@@ -73,9 +77,17 @@ export class Channel<Ctx> {
...
     this.emitter = new EventEmitter();
+    this.onUnrecoverableError = onUnrecoverableError;
   }
 
   public onCommand = (listener: (cmd: api.Command) => void) => {
+    if (this.closed) {
+      const e = new Error('Trying to listen to commands on a closed channel');
+      this.onUnrecoverableError(e);
+
+      throw e;
```
turbio, line 88: or not throw here?

src/channel.ts
```diff
@@ -101,14 +115,13 @@ export class Channel<Ctx> {
     });
 
     if (!this.sendToClient) {
-      throw new Error('Expected sendToClient');
+      this.onUnrecoverableError(new Error('Expected sendToClient'));
+
+      return;
```
turbio, line 120: hmmmm

turbio (approved):
:ok_hand: :warning:

src/channel.ts
masad-frost: Some functions are called by the client and I didn't wanna throw there, lemme get it to a bad state and see if I can just throw everywhere

masad-frost:
Always throwing seems to work, but depending on the error handling of the caller you might see the error once on your global error handler and once on the error callback.

masad-frost:
Both errors are the same instance, so for error reporting metrics we should be good.

PR #55, opened 2020-09-23
Simplify channel API and make connecting and closing more robust

masad-frost:
Why
===
The channel being responsible for closing adds a lot of complexity to the code. On the surface it looks like it simplifies the client but in reality it makes it harder to deal with and reason about closing and reconnecting. The channel also had unnecessary APIs that made the code harder and the closing API inconsistent.

There were a few bugs that (tests reflect those) caused opening, connecting, retrying and closing to go out of sync. The origin of those bugs is due to the fact that the channel 0 closing relied on the internals of the channel which was hard to keep track of.

Because of how things are intertwind right now this PR had to be larger than I'd liked.

What changed
============
On a higher level:
- Reduced the API surface area of channels
- Moved all channel opening and closing logic to the client
- Sprinkled asserts all over the place
- Documented internals more clearly

API changes:
- `channel.close` is no longer a thing. You can only close using the return function from `client.openChannel`
- Removed a bunch of unused public instance variables on channels
- Channel now has a status rather than `closed` boolean indicating the different states it can be in `open` | `closed` | `closing`
  - Changed asserts to use that status instead
- Channels can still receive commands after calling close, until the channel is actually closed (or the connection is closed)
- `openChannel` cannot be called before `open`
- `client.close` can be called at any time without errors and closes predictibly

**channel internals**
- Removed event emitter in favor of simple callback array
  - In turn, removed setMaxListeners and getMaxListeners
- Channels are not responsible for opening and closing, we still have a `handleClose` function called by the client to cleanup internals and set state

**client internals**
- A lot more asserts for state invariance
- Channel requests now can be in 2 states with a discriminant `isOpen`
- Channel requests are used for opening and closing and has the relevant information for that
- Channels are properly closed regardless of when the close request is sent. In the past we can end up in a nasty place if it's called while openning is inflight.
- Channels are now only created after we open the channel. Previously we put currentChannel on the request before actually opening
- Now that channels are not responsible for opening and closing, we now store the cleanup cb returned from `chan0Cb` and call it when necessary inside the client

### DIFF
--- a/src/__tests__/index.test.ts
+++ b/src/__tests__/index.test.ts
@@ -55,23 +55,59 @@ function genToken() {
 jest.setTimeout(10 * 1000);
 
 test('client connect', (done) => {
-  const onUnrecoverableError = jest.fn<void, [Error]>();
-  const client = new Client();
-  client.setUnrecoverErrorHandler(onUnrecoverableError);
+  const client = new Client<{ username: string }>();
+  client.setUnrecoverErrorHandler(done);
+
+  const ctx = { username: 'zyzz' };
 
   client.open(
     {
-      fetchToken: () => Promise.resolve({ token: REPL_TOKEN, aborted: false }),
+      fetchToken: () => Promise.resolve({ token: genToken(), aborted: false }),
       WebSocketClass: WebSocket,
+      context: ctx,
     },
-    ({ channel, error }) => {
-      expect(channel?.closed).toBe(false);
+    ({ channel, error, context }) => {
+      expect(channel?.status).toBe('open');
+      expect(context).toBe(ctx);
       expect(error).toEqual(null);
 
-      setTimeout(() => client.close());
+      client.close();
 
       return () => {
-        expect(onUnrecoverableError).toHaveBeenCalledTimes(0);
+        done();
+      };
+    },
+  );
+});
+
+test('client retries', (done) => {
+  const client = new Client();
+  client.setUnrecoverErrorHandler(done);
+
+  let tryCount = 0;
+
+  client.open(
+    {
+      fetchToken: () => {
+        tryCount += 1;
+
+        if (tryCount === 1) {
+          return Promise.resolve({ token: 'test - bad token retries', aborted: false });
+        }
+
+        return Promise.resolve({ token: genToken(), aborted: false });
+      },
+      WebSocketClass: WebSocket,
+      context: null,
+    },
+    ({ channel, error }) => {
+      expect(tryCount).toBe(2);
+      expect(channel?.status).toBe('open');
+      expect(error).toEqual(null);
+
+      client.close();
+
+      return () => {
         done();
       };
     },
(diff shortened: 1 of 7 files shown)


src/client.ts
```diff
@@ -570,52 +789,53 @@ export class Client<Ctx extends unknown = null> {
...
+          this.handleConnect();
 
-          // If user called close inside open callback (throws a error) we should not continue connecting
-          if (!closedDuringConnect) {
-            this.close = originalClose;
+          // defer closing if the user decides to call client.close inside chan0Cb
+          const originalClose = this.close;
+          this.close = () => setTimeout(() => {
+              originalClose();
+            }, 0);
```
sergeichestakov, line 820: I thought we wanted to throw if a user calls `close` inside the chan0cb?

sergeichestakov (commented):
Looking good! I'm definitely pro moving the complexity out of the Channel. Left some comments and can do another pass after

src/__tests__/index.test.ts
```diff
@@ -42,29 +119,29 @@ test('channel closing itself when client willReconnect', (done) => {
   let clientOpenCount = 0;
   let channelOpenCount = 0;
 
-  const onUnrecoverableError = jest.fn<void, [Error]>();
   const client = new Client();
-  client.setUnrecoverErrorHandler(onUnrecoverableError);
+  client.setUnrecoverErrorHandler(done);
```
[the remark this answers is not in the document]
masad-frost: Calling done with an error ends the tests and reports the error. The other done means the test passed. If any unrecoverable error does occur, we fail the test immediately

src/client.ts
```diff
@@ -154,37 +222,75 @@ export class Client<Ctx extends unknown = null> {
...
+   *
    */
   public openChannel = (options: ChannelOptions<Ctx>, cb: OpenChannelCb<Ctx>) => {
+    if (!this.chan0Cb) {
+      throw new Error('You must call client.open before attempting to open any channels');
+    }
+
+    if (options.name && this.channelRequests.some((cr) => cr.options.name === options.name)) {
+      throw new Error(`Channel with name ${options.name} already opened`);
+    }
```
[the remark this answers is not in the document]
masad-frost: Yes. Unnamed channels cannot conflict, you always end up creating a new channel to the service. If the same client tries to open a channel with the same name multiple times the container panics, this is to protect against that and start handling it on the client if it ever happens

src/client.ts
masad-frost: Unclear what the right move is here. Don't you think calling close in the callback is potentially something that we might wanna do?

masad-frost:
Appreciate it :pray:

src/__tests__/index.test.ts
```diff
@@ -1,36 +1,113 @@
...
+
+  return token.toString('base64');
 }
 
 jest.setTimeout(10 * 1000);
 
 test('client connect', (done) => {
-  const onUnrecoverableError = jest.fn<void, [Error]>();
+  const client = new Client<{ username: string }>();
+  client.setUnrecoverErrorHandler(done);
```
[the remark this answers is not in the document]
masad-frost: Oh yikes. I made a typo in the client as well.

cbrewster:
I'm still getting more familiar with crosis internals, but this looks like a solid improvement. Glad we are more explicitly tracking and enforcing the various states a channel can be in and it seems like this will remove some foot-guns from the public API (some of which I hit while writing the repl tester).

src/client.ts
```diff
@@ -284,13 +385,99 @@ export class Client<Ctx extends unknown = null> {
...
       this.channels[id] = channel;
-      channelRequest.currentChannel = channel;
+      // TODO we should stop relying on mutating the same channelrequest
+      (channelRequest as ChannelRequest<Ctx>).channelId = id;
+      (channelRequest as ChannelRequest<Ctx>).isOpen = true;
+      (channelRequest as ChannelRequest<Ctx>).cleanupCb = openChannelCb({
+        channel,
+        error: null,
+        context: this.connectOptions.context,
+      });
```
sergeichestakov, line 401: yeah mutating every request isn't ideal. Perhaps we should replace with a new channel request at that index but nbd

sergeichestakov (approved):
Those comments tho 😍 looks good to me

src/client.ts
masad-frost: Unfortunately I took the liberty of also relying on the reference here, in the case explained in this comment. https://github.com/replit/crosis/blob/b7922ae698d5b074534280191576713b38b1a56f/src/client.ts#L280-L281

Leading to this code path https://github.com/replit/crosis/blob/b7922ae698d5b074534280191576713b38b1a56f/src/client.ts#L403-L407. There are absolutely ways around it but it complicated things too much :(. Maybe I should turn this here into a code comment about how we rely on mutating the request

PR #56, opened 2020-10-05
[RFC] Allow client to re-use a websocket

masad-frost:
Why
===
We're currently in a place where we wanna rollout v6. Rolling out 2 websockets lead to some issues, such as having to request captcha more than once to get a token, increase load on the backend, 2x sessions per user, out of sync opening and closing of clients...etc

What changed
============
Added a new option to `open` args which is `getSocket`. It expects a promise that returns a websocket and that will be used to communicate with goval.

Test plan
=========
I wrote a simple test case, seems to work. I also tried to do it by using a socket provided by v6

### DIFF
--- a/src/__tests__/index.test.ts
+++ b/src/__tests__/index.test.ts
@@ -651,7 +651,7 @@ test('fetch abort signal works as expected', (done) => {
   );
 });
 
-test.only('client multiplexed', (done) => {
+test('client multiplexed', (done) => {
   const wsSourceClient = new Client();
   const clientMultiplexed = new Client();
 
--- a/src/__tests__/index.test.ts
+++ b/src/__tests__/index.test.ts
@@ -63,6 +63,7 @@ test('client connect', (done) => {
   client.open(
     {
       fetchToken: () => Promise.resolve({ token: genToken(), aborted: false }),
+      withPreconnectedSocket: false,
       WebSocketClass: WebSocket,
       context: ctx,
     },
@@ -97,6 +98,7 @@ test('client retries', (done) => {
 
         return Promise.resolve({ token: genToken(), aborted: false });
       },
+      withPreconnectedSocket: false,
       WebSocketClass: WebSocket,
       context: null,
     },
@@ -125,6 +127,7 @@ test('channel closing itself when client willReconnect', (done) => {
   client.open(
     {
       fetchToken: () => Promise.resolve({ token: genToken(), aborted: false }),
+      withPreconnectedSocket: false,
       WebSocketClass: WebSocket,
       context: null,
     },
@@ -184,6 +187,7 @@ test('channel open and close', (done) => {
   client.open(
     {
       fetchToken: () => Promise.resolve({ token: genToken(), aborted: false }),
+      withPreconnectedSocket: false,
       WebSocketClass: WebSocket,
       context: null,
     },
@@ -230,6 +234,7 @@ test('channel skips opening', (done) => {
   client.open(
     {
       fetchToken: () => Promise.resolve({ token: genToken(), aborted: false }),
+      withPreconnectedSocket: false,
       WebSocketClass: WebSocket,
       context: ctx,
     },
@@ -270,6 +275,7 @@ test('channel skips opening conditionally', (done) => {
   client.open(
     {
       fetchToken: () => Promise.resolve({ token: genToken(), aborted: false }),
+      withPreconnectedSocket: false,
       WebSocketClass: WebSocket,
       context: null,
     },
@@ -330,7 +336,6 @@ test('client rejects opening channel before client opens', () => {
   }).toThrow();
 });
 
-
 test('client rejects opening channel after closing client', () => {
   const client = new Client();
 
@@ -340,12 +345,12 @@ test('client rejects opening channel after closing client', () => {
   client.open(
     {
       fetchToken: () => Promise.resolve({ token: null, aborted: true }),
+      withPreconnectedSocket: false,
       WebSocketClass: WebSocket,
       context: null,
     },
     () => {},
   );
-
 
   client.close();
 
@@ -363,6 +368,7 @@ test('client rejects opening same channel twice', (done) => {
   client.open(
     {
       fetchToken: () => Promise.resolve({ token: genToken(), aborted: false }),
+      withPreconnectedSocket: false,
       WebSocketClass: WebSocket,
       context: null,
     },
@@ -386,7 +392,6 @@ test('client rejects opening same channel twice', (done) => {
   );
 });
 
-
 test('client reconnects unexpected disconnects', (done) => {
   const onUnrecoverableError = jest.fn<void, [Error]>();
   const client = new Client();
@@ -400,6 +405,7 @@ test('client reconnects unexpected disconnects', (done) => {
   client.open(
     {
       fetchToken: () => Promise.resolve({ token: genToken(), aborted: false }),
+      withPreconnectedSocket: false,
       WebSocketClass: WebSocket,
       context: null,
     },
@@ -471,6 +477,7 @@ test('client is closed while reconnecting', (done) => {
   client.open(
     {
       fetchToken,
+      withPreconnectedSocket: false,
       WebSocketClass: WebSocket,
       context: null,
     },
@@ -513,6 +520,7 @@ test('closing before ever connecting', (done) => {
 
         return Promise.resolve({ token: genToken(), aborted: false });
       },
+      withPreconnectedSocket: false,
       WebSocketClass: WebSocket,
       context: null,
     },
@@ -569,7 +577,8 @@ test('closing before ever connecting', (done) => {
 //   client.open(
 //     {
 //       fetchToken: () => Promise.resolve({ token: 'bad token', aborted: false }),
-//       WebSocketClass: WebSocket,
+//        withPreconnectedSocket: false,
+//        WebSocketClass: WebSocket,
 //       timeout: 0,
 //     },
 //     open,
@@ -592,6 +601,7 @@ test('fetch token fail', (done) => {
       fetchToken: () => {
         throw new Error('fail');
       },
+      withPreconnectedSocket: false,
       WebSocketClass: WebSocket,
       context: null,
     },
@@ -624,6 +634,7 @@ test('fetch abort signal works as expected', (done) => {
             client.close();
           }, 0);
         }),
+      withPreconnectedSocket: false,
       WebSocketClass: WebSocket,
       context: null,
     },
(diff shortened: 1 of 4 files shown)


src/__tests__/index.test.ts
```diff
@@ -639,3 +650,72 @@ test('fetch abort signal works as expected', (done) => {
...
+      WebSocketClass: WebSocket,
+      context: null,
+    },
+    ({ channel, error }) => {
+      expect(channel?.status).toBe('open');
+      expect(error).toBeNull();
+
+      // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+      // @ts-ignore (ws is private)
+      const { ws } = wsSourceClient;
```
sergeichestakov, line 680: perhaps we should just expose a `client.getWS()` method since we're reaching in to grab the WS so much anyways and presumably that's what you'd need to do to reuse a WS in this way.

src/types.ts
```diff
@@ -40,10 +40,28 @@ export interface UrlOptions {
...
   fetchToken: (abortSignal: AbortSignal) => Promise<{ token: null, aborted: true } | { token: string, aborted: false }>;
+  context: Ctx;
   urlOptions: UrlOptions;
   timeout: number | null;
   WebSocketClass?: typeof WebSocket;
+}
+
+interface ConnectOptionsWithSocket<Ctx> {
+  withPreconnectedSocket: true,
+  getSocket: (abortSignal: AbortSignal) => Promise<{ ws: null, aborted: true } | { ws: WebSocket, aborted: false }>;
```
sergeichestakov, line 54: seems redundant to have both of these fields since they're always used in conjunction and a defined `getSocket` function implies `withPreconnectedSocket` is true

src/__tests__/index.test.ts
```diff
@@ -639,3 +650,72 @@ test('fetch abort signal works as expected', (done) => {
     },
   );
 });
+
+test.only('client multiplexed', (done) => {
```
sergeichestakov, line 654: why `test.only`? don't we want the other tests to run too

sergeichestakov (commented):
While I understand the motivation behind this PR, isn't the ideal scenario that we only use one client to begin with? Seems like it'd be better to tackle the root cause of the issue of opening two Websocket connections rather than work around it

src/__tests__/index.test.ts
masad-frost: Whoops, my bad

src/types.ts
masad-frost: I think we've discussed this in other contexts. When we have a union type like this it's always better to have a clear discriminant

src/__tests__/index.test.ts
masad-frost: Yeah good point

masad-frost:
I think we've discussed this in detail multiple times and this seemed the lowest friction path towards v6.

Our options are:

- rewrite everything using v6
- write gluecode and adapters for v5 to v6
- slowly migrate to v6 and services by multiplexing the socket

This seems like our safest and fastest path to v6

src/__tests__/index.test.ts
masad-frost:
```suggestion
test('client multiplexed', (done) => {
```

masad-frost:
That is correct.

We're not actually multiplexing the socket, we're just reusing it. As far as the backend is concerned it's 1 client.

src/client.ts
```diff
@@ -197,19 +190,29 @@ export class Client<Ctx extends unknown = null> {
       throw new Error('You must call `close` before opening the client again');
     }
 
-    if (!options.fetchToken || typeof options.fetchToken !== 'function') {
-      throw new Error('You must provide a fetchToken function');
-    }
+    if (options.withPreconnectedSocket) {
+      if (!options.getSocket || typeof options.getSocket !== 'function') {
```
dan-stowell, line 194: Naive typescript question: is it possible to enforce `getSocket` being a function at compile-time?

src/client.ts
```diff
@@ -197,19 +190,29 @@ export class Client<Ctx extends unknown = null> {
...
+      }
+
+      this.connectOptions = {
+        timeout: 10000,
+        urlOptions: {
+          secure: false,
+          host: 'eval.repl.it',
+          port: '80',
+        },
+        ...options,
```
dan-stowell, line 213: More n00b questions: if someone passes in `timeout` or `urlOptions` in `options`, will those override what's set here?

src/client.ts
```diff
@@ -604,16 +608,110 @@ export class Client<Ctx extends unknown = null> {
...
+        this.onUnrecoverableError(e);
+
+        return;
+      }
+
+      this.tokenOrSocketAbortController = null;
+
+      const { ws, aborted } = socketGetterResult;
+
+      if (abortController.signal.aborted !== aborted) {
```
dan-stowell, line 634: Where is the `AbortController` interface defined?

src/client.ts
```diff
@@ -604,16 +608,110 @@ export class Client<Ctx extends unknown = null> {
...
+        // the aborted return value and the abort signal should be equivalent
+        if (abortController.signal.aborted) {
+          // In cases where our abort signal has been called means `client.close` was called
```
dan-stowell, line 637: Can we assert this is the case?

dan-stowell: Nevermind, I'm picking up the vibe that calling `onUnrecoverableError` and throwing **are** our way of asserting.

src/client.ts
```diff
@@ -604,16 +608,110 @@ export class Client<Ctx extends unknown = null> {
...
+
+        return;
+      }
+
+      this.ws = ws;
+      ws.onmessage = this.onSocketMessage;
+
+      this.handleConnect();
+
+      // defer closing if the user decides to call client.close inside chan0Cb
```
dan-stowell, line 699: Do we ever need to call `client.close()` in `chan0Cb`? Or are we just keeping the option open?

dan-stowell: Feels sketchy to me to reassign `this.close` like this.

dan-stowell (changes requested):
Thank you for plunging in and offering up this RFC, @masad-frost ! I'm on board with the high-level approach of using a single socket across clients. I have a few questions because I'm a crosis and TypeScript n00b, thanks in advance for enlightening me (and anyone else reading this PR).

Just by reading this PR, I can't tell whether client connection errors will go up, stay the same, or go down. If they go up, is it because we have written more checks into `open()` and `connect()` or because we've introduced bugs? What would give me a lot of confidence is a state diagram and tests for all the transitions in the state diagram. Right now I think the state we're keeping is:

* are we connected or disconnected?
* has the connection been aborted?
* has channel 0 been closed?

This is less important than testing and getting us down to one connection in a reasonable timeframe, but something to think about: can we represent states in the state diagram as types? So we can do a type check to know which state we're in? And, possibly, in the future, disallow certain operations on certain types? Enums also suffice!

The dance around `chan0Cb` possibly calling `close()` on the connection seems fraught to me - do we have to support that?

turbio (changes requested):
I'm finding the code a bit hard to understand, I'm less concerned with the implementation and more with the design. If I'm understanding correctly we plan on having two clients (v5 and v6) share the same websocket. Is there somewhere I can find a good description of how we expect two separate clients to cooperatively share one websocket? This sounds really hard so I'm suspecting I'm misunderstanding.

From your comment

> I think we've discussed this in detail multiple times and this seemed the lowest friction path towards v6.
> 
> Our options are:
> 
>     rewrite everything using v6
>     write gluecode and adapters for v5 to v6
>     slowly migrate to v6 and services by multiplexing the socket
> 
> This seems like our safest and fastest path to v6

I'm **guessing** from context we're going with the last option. But I'm really unclear about how that last option will work. Could you write some more about what the plan is?

masad-frost:
> Just by reading this PR, I can't tell whether client connection errors will go up, stay the same, or go down. If they go up, is it because we have written more checks into open() and connect() or because we've introduced bugs? What would give me a lot of confidence is a state diagram and tests for all the transitions in the state diagram. 

Here is expected state after:
1. `new Client`
    - `connectionState` is DISCONNECTED
    - `ws` is null
    - `connectOptions` is null
    - `chan0Cb` is null
    - `chan0CleanupCb` is null
    - `channelRequests` is empty
    - `channels` is empty
    - `retryTimeoutId` is null
    - `tokenOrSocketAbortController` is null
2. `client.open` which calls `connect()` synchronously
    - `connectionState` is CONNECTING
    - `ws` is null
    - `connectOptions` is whatever the user supplied as a second argument (with some defaults)
    - `chan0Cb` is whatever the user supplied as a second argument
    - `chan0CleanupCb` is null
    - `channelRequests` is empty
    - `channels` should have a single channel (channel 0)
    - `retryTimeoutId` is null
    - `tokenOrSocketAbortController` is an AbortController
3. After the token is successfully fetched (assuming client.close wasn't called)
    - `connectionState` is CONNECTING
    - `ws` is a websocket instance
    - `connectOptions` is whatever the user supplied as a second argument (with some defaults)
    - `chan0Cb` is whatever the user supplied as a second argument
    - `chan0CleanupCb` is null
    - `channelRequests` depends on the calls to `client.openChannel`
    - `channels` should have a single channel (channel 0)
    - `retryTimeoutId` is null
    - `tokenOrSocketAbortController` is null
4. Socket closes or we hit the timeout limit while trying to establish a handshake with pid1 (we will keep retrying until `client.close` is called)
    - `connectionState` is CONNECTING
    - `ws` is null
    - `connectOptions` is whatever the user supplied as a second argument (with some defaults)
    - `chan0Cb` is whatever the user supplied as a second argument
    - `chan0CleanupCb` is null
    - `channelRequests` depends on the calls to `client.openChannel`
    - `channels` should have a single channel (channel 0)
    - `retryTimeoutId` is a timeout id scheduling a retry
    - `tokenOrSocketAbortController` is null
5. Connection established
    - `connectionState` is CONNECTED
    - `ws` is a websocket instance
    - `connectOptions` is whatever the user supplied as a second argument (with some defaults)
    - `chan0Cb` is whatever the user supplied as a second argument
    - `chan0CleanupCb` is the returned function from calling chan0Cb (could be undefined)
    - `channelRequests` depends on the calls to `client.openChannel`
    - `channels` should have any channel that is currently open
    - `retryTimeoutId` is null
    - `tokenOrSocketAbortController` is null
6. Also any time we call `client.close()`, we should get be back to the initial state, regardless of the state we're in

We do have tests for the different states we can be in, we don't explicitly check internal state but there's a buttload of state invariance checks inside the client so the tests will automatically fail. Maybe I can codify these states more explicitly, but i feel like this is not relevant to this PR, we can discuss in a separate thread. 

In the case of supplying a websocket, the websocket should already be established through another client and got container `READY` already. So in essence there cannot be any connection errors.

masad-frost:
> I'm guessing from context we're going with the last option. But I'm really unclear about how that last option will work. Could you write some more about what the plan is?

Sure can. The idea is that the current client we use (v5) will be supplying a socket that is connected and did the handshake with PID1. Once we call `clientv6.open` we will pass a `getSocket` function that will get the socket from the v5 client and supply it to the v6 client. Coordination is not a problem as both will connect and disconnect at the same time (they're using the same socket), the only part we have to worry about is making sure we give clientv6 a socket from a connected clientv5.

src/client.ts
masad-frost: Yeah, it is already the case (see `ConnectArgs`), this is an extra runtime check since at the end of the day everything will be running in javascript.

src/client.ts
masad-frost: Yes.
```js
var overrides = { wat: false }
console.log({  wat: true, ...overrides }) // logs { wat: false }
console.log({ ...overrides, wat: true }) // logs { wat: true }
```

src/client.ts
masad-frost: https://developer.mozilla.org/en-US/docs/Web/API/AbortController

src/client.ts
masad-frost: Yes `onUnrecoverableError` is crash and die.

src/client.ts
masad-frost: I personally use it heavily in tests so it's more of a convenience. It does feel wrong, but it's the javascript way :rofl:, we reassign it synchronously so it's fine. What we can do instead is make it an error to call close inside `chan0Cb` and move the `setTimeout` requirement to the caller

masad-frost:
Lemme know if you guys have further comments, otherwise, there are some nitty gritties i need to fix before needing a final review

masad-frost:
If there are no further comments, I'm gonna be zooming in on this later today

dan-stowell:
Thank you for writing out the state diagram. Given that and the plan I can say this code executes both. My one question, which can talk about off-pull-request, is what's pressuring us to get to crosis v6? Bug fixes? Supporting services on the new anon language page?

masad-frost:
Moudy and I think it's a big API improvement based on what we've been doing
in the workspace. Yes we're already using it in the new anon page and the
social view. We also built a bunch of services that are well tested and
have a nice API. We could've built services with v5 but it would've been
unergonomic

dan-stowell:
@masad-frost and I talked about services yesterday. We're going to ship a recent version of v6 for Moudy to use with the anonymous language page, and then pick the highest-value service to bring into the workspace. Then we'll revisit this PR.

masad-frost:
We ended up with 2 websockets and we think it's ok

PR #57, opened 2020-10-15
Allow openning channels before client connects

masad-frost:
Why
===

After some thinking and discussing, seems like it is sane to allow `client.openChannel` to be called at any point in the lifecycle of the client.

What changed
============

- Removed assert
- Replaced a couple of tests

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -242,10 +242,6 @@ export class Client<Ctx extends unknown = null> {
    *
    */
   public openChannel = (options: ChannelOptions<Ctx>, cb: OpenChannelCb<Ctx>) => {
-    if (!this.chan0Cb) {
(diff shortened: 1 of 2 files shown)


src/client.ts
```diff
@@ -242,10 +242,6 @@ export class Client<Ctx extends unknown = null> {
    *
    */
   public openChannel = (options: ChannelOptions<Ctx>, cb: OpenChannelCb<Ctx>) => {
-    if (!this.chan0Cb) {
```
moudy, line 245: did this ever get used ii the app of did we not upgrade yet?

moudy (approved):

src/client.ts
masad-frost: We haven't upgraded yet.

PR #58, opened 2020-10-15
Maintain openChannel requests when the client closes

masad-frost:
`Why
===

Somewhat adjacent to #57. We think this is much more ergonomic and it allows people to use the client with more ease. A usecase for us is that we don't want to reinitialize services and call `openChannel` again when we have to forcefully reconnect (i.e. fork or repl slug change), we'd just have to call `client.close` and then `client.open`, all the channels will be ready to open.

What
============
- Even if the client isn't automatically reconnecting we keep the channel requests around for future manual reconnects (through client.open)
- Added a `destroy` function that cleans up channelRequests and some callbacks to avoid any potential leaks, it also puts the client in a state where trying to reconnect it throws.
  - Ideally we just lose reference to the root object anyway, but this is just to be safe and to make sure we don't do something stupid after we're done with the client.

### DIFF
--- a/src/__tests__/index.test.ts
+++ b/src/__tests__/index.test.ts
@@ -356,6 +356,50 @@ test('client rejects opening channel after closing client', () => {
   expect(errorHandler).toHaveBeenCalledTimes(0);
 });
 
+
+test('closing maintains openChannel requests', (done) => {
+  const client = new Client();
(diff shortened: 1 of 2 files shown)


moudy (approved):
🎖️

PR #59, opened 2020-10-15
[v5] Assert named channel isn't opened twice

masad-frost:
Why
===
Opening a channel with the same name multiple times causes the container to panic and die. Let's make sure we catch it early on the client so that not all users are affected by a single client, and it'll give us a better idea of why it's happening since we'll have client side traces that reference the error.

What changed
============
- Assert open channels don't have the same name as the new open channel request
- Same but for channels that are inflight and awaiting to be open

Test plan
=========
- Open 2 channels with the same name
- Second call throws

### DIFF
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -16,6 +16,8 @@ export class Channel extends EventEmitter {
 
   public closed: boolean;
 
+  public name: string | null
+
   // private
   private sendQueue: Array<api.ICommand>;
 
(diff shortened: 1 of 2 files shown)


src/channel.ts
```diff
@@ -16,14 +16,16 @@ export class Channel extends EventEmitter {
 
   public closed: boolean;
 
+  public name: string | null
```
turbio, line 19: should it be public?

turbio (approved):
thnx! 🚂

src/channel.ts
masad-frost: meh

masad-frost: Oh actually it should be for us to check in the client :P

PR #60, opened 2020-10-20
Fix client initializing destroyed

masad-frost:
Why
===

The client should only switch to be beying destroyed after `destroy` has been called. This is a stupid bug, must've slipped in while testing, we need CI on this repo YIKES.

What changed
============

Flipped destroyed flag to start `false`

Test plan
=========
before 
image
after
image


moudy (approved):
💥

PR #62, opened 2020-11-12
Channel closing improvements

masad-frost:
Why
===
You cannot call closeChannel multiple times right now. It's really hard to tell whether or not you can call it. It's also hard to tell  whether you need to call it or not. There's no really good point in limiting the number of calls, so let's allow it in case someone needs to make sure the channel closes and doesn't re-open.

What changed
============
- Allow calling `closeChannel` multiple times
- Don't return a promise from `closeChannel`
  - Although it would be cool to know when the channel actually closes after calling close, we don't have a good reason to need that right now. Now with multiple calls the returned promise will not be honest unless we save the promise or something nasty. Even before this change there was a chance that the opening is inflight and there's a future promise that should've been returned, see: https://github.com/replit/crosis/blob/b1d44baee2d40a21c2f3b7bd1c28a38d94401552/src/client.ts#L431-L435
- Previously if the client disconnects (chan0 closes) during a request to close a channel we returned early from the `requestCloseChannel`. That doesn't make much sense. If the client reconnects we end up opening the channel then requesting a close again lol. Now we keep going with the cleanup regardless of whether the client disconnected during the request or not. 
  - Improved breadcrumbs and asserts around this

Test plan
=========
None of the tests broke :heavy_check_mark: 
If you see a test that you think should be added, lmk

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -441,6 +441,8 @@ export class Client<Ctx extends unknown = null> {
       return;
     }
 
+    const { channelId } = channelRequest;
+
     const chan = this.getChannel(channelRequest.channelId);
     chan.status = 'closing';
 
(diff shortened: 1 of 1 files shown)


moudy (approved):
📕

PR #63, opened 2020-11-18
Support fetching connection metadata instead of just the token

lhchavez:
Why
===

This is part of a migration that allows all the information that's needed to construct the URL to be provided by a single function.

What changed
============

This change starts the deprecation of `fetchToken`/`urlOptions`
combination as what gets the connection token / URL and instead relies
on `fetchConnectionMetadata`, which does both.

Test plan
=========

```shell
yarn test
```

(this also has a backwards-compatibility layer to avoid breaking existing clients).

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -1,7 +1,7 @@
 import { api } from '@replit/protocol';
 import { Channel } from './channel';
 import { getWebSocketClass, getNextRetryDelay, getConnectionStr } from './util/helpers';
-import { ConnectOptions, ClientCloseReason, ChannelCloseReason, ChannelOptions } from './types';
+import { ConnectOptions, ClientCloseReason, ChannelCloseReason, ChannelOptions, UrlOptions } from './types';
 
 /**
  * The only required option is `fetchConnectionMetadata` (falling back to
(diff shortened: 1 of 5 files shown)


src/client.ts
```diff
@@ -220,12 +252,9 @@ export class Client<Ctx extends unknown = null> {
...
 
     this.connectOptions = {
+      fetchConnectionMetadata,
       timeout: 10000,
-      urlOptions: {
-        secure: false,
-        host: 'eval.repl.it',
-        port: '80',
-      },
+      urlOptions: defaultUrlOptions,
```
masad-frost, line 257: Doesn't seem like we need this in connectOptions anymore, only in the arg

src/client.ts
```diff
@@ -203,12 +215,32 @@ export class Client<Ctx extends unknown = null> {
...
+        return {
+          connectionMetadata: {
+            token,
+            gurl: `ws${secure ? 's' : ''}://${host}:${port}`,
+            conmanURL: `http${secure ? 's' : ''}://${host}:${port}`,
+          },
+          aborted: false,
+        };
+      };
     }
```
masad-frost, line 244: I like it!

masad-frost (approved):
:egg:cellent!

src/client.ts
lhchavez: oh yeah! moved it to the args.

PR #64, opened 2020-11-19
Support connection metadata instead of just the token

lhchavez:
Why
===

This is part of a migration that allows all the information that's needed to construct the URL to be provided by a single function. Crosisv5 edition!

What changed
============

This change starts the deprecation of the `token`/`urlOptions` combination
in favor of `connectionMetadata`, which contains all the information
needed.

Test plan
=========

localhost repl.it still works!

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -1,6 +1,7 @@
 /* global WebSocket */
 
 import { EventEmitter } from 'events';
+import * as urllib from 'url';
 import { api } from '@replit/protocol';
 import { Channel } from './channel';
 import { EIOCompat } from './EIOCompat';
@@ -12,10 +13,16 @@ enum ConnectionState {
   DISCONNECTED = 2,
 }
 
-interface UrlOptions {
+export interface UrlOptions {
   secure: boolean;
   host: string;
   port: string;
+}
+
+interface GovalMetadata {
+  token: string;
+  gurl: string;
+  conmanURL: string;
 }
 
 interface TxRx {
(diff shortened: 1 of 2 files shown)


src/client.ts
```diff
@@ -39,13 +46,32 @@ type DebugLog =
...
+  // WebSockets are not supported.
+  pollingHost?: string;
+
+  timeout: number | null;
+  WebSocketClass: typeof WebSocket;
+}
+
+export interface ConnectArgs extends Partial<ConnectOptions> {
+  // token, urlOptions, and polling have been deprecated in favor of
+  // connectionMetadata.
```
masad-frost, line 61: :pray:

masad-frost (approved):
LGTM!

Few notes about polling fallback:

- It always uses eval.repl.it, so that needs to be fixed
- Code is here https://github.com/replit/goval-proxy-v2
- There's a graph at the bottom of this dashboard indicating how many times it happens https://repl-it-metabase.herokuapp.com/dashboard/41
- It has prometheus metrics, but I can't seem to find a grafana dashboard for it :/. Hope prom is pulling it in.
- I think the majority of the fallbacks are actually due to blips in goval itself rather than clients not supporting websockets, we haven't really bothered investigating. 

LMK if you need help with it.

lhchavez:
> LGTM!
> 
> Few notes about polling fallback:
> 
> * It always uses eval.repl.it, so that needs to be fixed
> * Code is here https://github.com/replit/goval-proxy-v2
> * There's a graph at the bottom of this dashboard indicating how many times it happens https://repl-it-metabase.herokuapp.com/dashboard/41
> * It has prometheus metrics, but I can't seem to find a grafana dashboard for it :/. Hope prom is pulling it in.
> * I think the majority of the fallbacks are actually due to blips in goval itself rather than clients not supporting websockets, we haven't really bothered investigating.
> 
> LMK if you need help with it.

awesome, i was going to ask about that soon-ish because that's going to be what i fix after everything in repl-it-web is Lore-compatible. thanks!

PR #69, opened 2020-11-27
Setup actions

masad-frost:
We're setting up actions. It works, we'll follow up for fixes to the lint problem and figure out how we're gonna run tests on the v6 branch with secret keys and all.


Scoder12:
#69 nice

masad-frost:
nice

firefish111:
very nice

PR #74, opened 2020-12-11
Tweak the fetchConnectionMetadata's result interface a bit

lhchavez:
Why
===

@sergeichestakov suggested in https://github.com/replit/repl-it-web/pull/8536#discussion_r541132620 some better names for the result fields. additionally, suggested to follow the typical type pattern used in repl-it-web: https://github.com/replit/repl-it-web/blob/149f4a6cd7d8b635826ab4d75bbbe9c0874e5dc4/app/services/fs/types.ts#L213-L215

What changed
============

This change makes the fetchConnectionMetadata return an object that is
compatible with the pattern that is used in repl-it-web.

Test plan
=========

Tests passed!

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -46,34 +46,28 @@ export interface GovalMetadata {
   conmanURL: string;
 }
 
-export enum FetchConnectionMetadataResult {
-  /**
-   * Fetch was successful.
-   */
-  Ok = 'Ok',
-
+export const FetchConnectionMetadataError = {
   /**
    * Fetch was aborted.
    */
-  Aborted = 'Aborted',
+  Aborted: new Error('Aborted'),
 
   /**
    * The fetch failed due to a recoverable error (mostly a transient network
    * condition).
    */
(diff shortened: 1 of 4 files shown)


src/types.ts
```diff
@@ -46,34 +46,28 @@ export interface GovalMetadata {
...
-   * Fetch was successful.
-   */
-  Ok = 'Ok',
-
+export const FetchConnectionMetadataError = {
   /**
    * Fetch was aborted.
    */
-  Aborted = 'Aborted',
+  Aborted: new Error('Aborted'),
```
masad-frost, line 53: Probably shouldn't construct an error statically like this. You can keep it as a string, though overall I think I liked the old format more.

masad-frost: I guess not statically, but I meant at the module.

lhchavez: yeah, i was a bit torn between consistency and type purity. back to string-based enums it is!

masad-frost (approved):
:1st_place_medal:

PR #75, opened 2020-12-11
Fix CI

lhchavez:
Why
===

We don't have CI!

What changed
============

Now with more v6 support!

Test plan
=========

```shell
yarn run lint
yarn run test
```


.github/workflows/main.yml
```diff
@@ -0,0 +1,33 @@
...
+      - uses: bahmutov/npm-install@v1
+
+      - name: Build
+        run: yarn run build
+
+      - name: Lint
+        run: yarn run lint
+
+      - name: Test
+        continue-on-error: true # Don't have a way of providing the token right now.
```
masad-frost, line 32: Seems like adding a secret is pretty straight forward https://docs.github.com/en/free-pro-team@latest/actions/reference/encrypted-secrets#using-encrypted-secrets-in-a-workflow

src/__tests__/index.test.ts
```diff
@@ -317,7 +318,7 @@ test('channel open and close', (done) => {
   });
 });
 
-test('channel skips opening', (done) => {
+test.skip('channel skips opening', (done) => {
```
masad-frost, line 321: This test seems to work fine

masad-frost (approved):
LGTM, let's add the secret and run the tests!

I'll send a PR to fix the other tests

masad-frost:
Sent a PR to fix tests https://github.com/replit/crosis/pull/76

.github/workflows/main.yml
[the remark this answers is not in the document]
lhchavez: oh right, these are available if the PR comes from the same repository! (in most of the other repos I've used secrets I run into the situation where PRs don't get access to these since they come from a fork)

done.

src/__tests__/index.test.ts
lhchavez: huh, i guess it does. it failed for me last week, but it's passing now, so no reason to skip it any longer.

PR #76, opened 2020-12-14
Fix a couple of tests (and merge #77)

masad-frost:
Also ended up merging #77 into this

Why
===

In #75 noticed some tests are skipped because they're broken.

What changed
============

- I'm not sure how `'client rejects opening same channel twice'` ever worked, but now that we don't need to open the client to open channels the test is straight forward, we just call open channel twice and expect the second call to throw.
- `'channel open and close'` was calling `closeChannel` synchronously from within `closeChanCb`, the client doesn't know how to handle it so I'm not sure how this test ever worked. (Sending a follow up to allow and handle this)
- `'channel skips opening'` seems to work fine

Test plan
=========
tests pass :)


src/__tests__/index.test.ts
```diff
@@ -304,7 +304,9 @@ test.skip('channel open and close', (done) => {
     expect(channel?.status).toBe('open');
     expect(error).toBe(null);
 
-    close();
+    setTimeout(() => {
+      close();
+    })
 
     expect(channel?.status).toBe('closing');
```
masad-frost, line 311:
```suggestion
    setTimeout(() => {
      close();
      expect(channel?.status).toBe('closing');
    });
```

sergeichestakov (approved):
Sweet

PR #77, opened 2020-12-14
Handle closing channel synchronously from within openChannelCb

masad-frost:
Why
===
The client has code to handle closing a channel after it's been opened. When the user calls close on a channel synchronously inside `openChannelCb` we end up in a weird state where we end up calling `requestCloseChannel` twice.

What changed
============

We might be able to clean this up and make it more robust later, but for now this works. We save the `closeRequested` flag before calling `openChannelCb` making it so that we only call `requestCloseChannel` once.

Test plan
=========
- Added a test

### DIFF
--- a/src/__tests__/index.test.ts
+++ b/src/__tests__/index.test.ts
@@ -306,7 +306,55 @@ test('channel open and close', (done) => {
 
     setTimeout(() => {
       close();
-    })
+    });
+
+    expect(channel?.status).toBe('closing');
+
+    return ({ willReconnect }) => {
+      expect(willReconnect).toBeFalsy();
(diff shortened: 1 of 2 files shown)


masad-frost:
Praise asserts for finding this bug!

src/__tests__/index.test.ts
```diff
@@ -309,6 +309,55 @@ test('channel open and close', (done) => {
       expect(channel?.status).toBe('closing');
     });
 
+
```
masad-frost, line 312:
```suggestion
```

moudy (approved):

PR #81, opened 2020-12-29
Add some breadcrumbs to figure out chan0CleanupCb error

masad-frost:
Why
===

Can't really figure out why this error is happening in production. The only thing that seems obvious is that it's related to timeouts. I've wrote tests that simulate timeouts to no avail will follow up with a PR to add those.

See sentry:
https://sentry.repl.it/repl-it/frontend/issues/219257

What changed
============
Added some breadcrumbs and removed noisey getchannel breadcrumb

Test plan
=========
:shrug:


sergeichestakov (approved):
🍞

PR #82, opened 2021-01-12
Implement polling fallback

masad-frost:
Why
===

It is very common for firewalls (especially school firewalls) to block websockets. Some of these requests go through to our infra and we notice a lot of `failed to upgrade to ws` errors.

What changed
============
- First thing I had to make EIOCompat work in node so i can test it easily.
  - Moved `binaryType` to setter and getter (in the past we were not setting it on the eio socket and it would shit itself in Nodejs)
  - MessageEvent, CloseEvent, and Event are not available in nodejs so we create a normal object instead
- `getConnectionStr` now returns a string with hardcoded polling server if passed `isPolling` flag.
- And for the intended change, if we try to connect and we never receive a (decodable) message on the websocket 3 times in a row, we switch to polling.
  - I decided to not add any options here but we can always add them retroactively

### DIFF
--- a/src/__tests__/index.test.ts
+++ b/src/__tests__/index.test.ts
@@ -2,6 +2,7 @@
 
 import type { FetchConnectionMetadataResult } from '../types';
 import { Client, FetchConnectionMetadataError } from '..';
+import { getWebSocketClass } from '../util/helpers';
 
 // eslint-disable-next-line
 const genConnectionMetadata = require('../../debug/genConnectionMetadata');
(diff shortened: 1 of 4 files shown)


src/client.ts
```diff
@@ -858,7 +869,19 @@ export class Client<Ctx extends unknown = null> {
...
-    const connStr = getConnectionStr(connectionMetadata);
+    if (websocketFailureCount === 3) {
+      // Report that we fellback to polling
+      this.debug({
+        type: 'breadcrumb',
+        message: 'polling fallback',
+      });
+    }
+
+    const isPolling = websocketFailureCount >= 3;
```
sergeichestakov, line 880: wasn't this configurable? should probably make it a connect option but we can follow up with that

src/__tests__/index.test.ts
```diff
@@ -731,54 +732,51 @@ test('closing before ever connecting', (done) => {
...
+          error: null,
+        }),
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore
+      WebSocketClass: WebsocketThatNeverConnects,
+      context: null,
+    },
+    ({ channel }) => {
+      expect(channel).toBeTruthy();
+      expect(didLogFallback).toBeTruthy();
```
sergeichestakov, line 773: nit: maybe we can be more specific here? e.g. 
```suggestion
    ({ channel, error }) => {
      expect(error).toBeNull();
      expect(channel).not.toBeNull();
      expect(didLogFallback).toBe(true);
```

src/__tests__/index.test.ts
```diff
@@ -731,54 +732,51 @@ test('closing before ever connecting', (done) => {
...
+  const onUnrecoverableError = jest.fn<void, [Error]>();
+  const client = new Client();
+  client.setUnrecoverableErrorHandler(onUnrecoverableError);
+
+  class WebsocketThatNeverConnects {
+    static OPEN = 1;
+  }
+
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
```
sergeichestakov, line 745: this is annoying - thoughts on updating the eslint rule for this?

sergeichestakov (approved):
Looks good!

lhchavez (approved):

masad-frost:
Looks like the CI repl is RIP, I'm not sure where it went lol

src/client.ts
masad-frost: It was configurable but this logic was outside of the client. We can think about it later, the only piece of config that probably fits here is to configure the number of failures before connecting.

src/__tests__/index.test.ts
masad-frost: :shrug:

masad-frost:
Can confirm all tests pass locally

PR #84, opened 2021-03-11
Add Client.getConnectionMetadata()

lhchavez:
Why
===

We want to be able to get the token without jumping through too many hoops, and v5 used to contain a similar mechanism to get the token.

What changed
============

This change adds a way of obtaining the connection metadata for the
WebSocket. This allows obtaining the connection's token, among other
things.

Test plan
=========

Call the getter.


masad-frost (approved):
:information_source:

PR #85, opened 2021-03-15
Allow passing a thunk for service when opening a channel

masad-frost:
Why
===

Sometimes we need some data from the context to decide what service to open (i.e. repl ownership information to decide `gcsfiles` vs `files`).

What changed
============
What the title says, basically we can pass a thunk that takes the context as an argument and return the service string that we need.

In the client we call the function or just use the passed in string conditionally.

Test plan
=========
- Added a test

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -109,12 +109,16 @@ export type ChannelCloseReason =
       willReconnect: false;
     };
 
+interface ServiceThunk<Ctx> {
+  (context: Ctx): string;
+}
+
 /**
  * See [[Client.openChannel]]
  */
 export interface ChannelOptions<Ctx> {
   name?: string;
-  service: string;
+  service: string | ServiceThunk<Ctx>;
   action?: api.OpenChannel.Action;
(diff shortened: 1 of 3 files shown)


src/types.ts
```diff
@@ -109,12 +109,16 @@ export type ChannelCloseReason =
       willReconnect: false;
     };
 
+interface ServiceThunk<Ctx> {
```
lhchavez, line 112: maybe add a nice docstring explaining that the return value is expected to be the service name, since the rest of the interface seems to be well-documented.

lhchavez (approved):
image

src/types.ts
masad-frost: For sure, thanks

masad-frost:
For reference, in the webapp I'm gonna use it here https://github.com/replit/repl-it-web/blob/master/app/services/fs/index.ts#L261-L267, right now we're using some sort of global.

moudy (approved):
nice

PR #86, opened 2021-03-15
Fix token generation for CI and debug mode

masad-frost:
Why
===
Tests can't run because we're using an old token generation method

What changed
============
Use shiny new token generation method. This hardcodes `global` cluster, we can improve it later on to contact lore if we want to. I added secrets to the repo, feel free to review those too https://docs.github.com/en/actions/reference/encrypted-secrets

Test plan
=========
Once goval updates with the new public key for crosis ci, the tests in here should run (and hopefully pass).
Keys added in replit/goval#2336

### DIFF
--- a/debug/genConnectionMetadata.js
+++ b/debug/genConnectionMetadata.js
@@ -1,9 +1,12 @@
 /* eslint-disable  @typescript-eslint/no-var-requires */
+const crypto = require('crypto');
 const { api } = require('@replit/protocol');
 const paseto = require('./paseto');
 
 const keyId = process.env.USER_KEY_ID;
-const govalPrivateKey = crypto.createPrivateKey(process.env.USER_PRIVATE_KEY_PEM.replace(/\\n/, '\n'));
+const govalPrivateKey = crypto.createPrivateKey(
+  process.env.USER_PRIVATE_KEY_PEM.replace(/\\n/, '\n'),
+);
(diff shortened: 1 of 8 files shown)


debug/genConnectionMetadata.js
```diff
@@ -1,53 +1,59 @@
 /* eslint-disable  @typescript-eslint/no-var-requires */
-const crypto = require('crypto');
+const { api } = require('@replit/protocol');
+const paseto = require('./paseto');
 
-const { TOKEN_SECRET } = process.env;
+const keyId = process.env.USER_KEY_ID;
+const govalPrivateKey = process.env.USER_PRIVATE_KEY_PEM.replace(/\\n/, '\n');
+const govalPublicKey = process.env.USER_PUBLIC_KEY_PEM.replace(/\\n/, '\n');
```
lhchavez, line 7:
```suggestion
const govalPrivateKey = crypto.createPrivateKey(process.env.USER_PRIVATE_KEY_PEM.replace(/\\n/, '\n'));
const govalPublicKey = crypto.createPublicKey(process.env.USER_PUBLIC_KEY_PEM.replace(/\\n/, '\n'));
```

the paseto implementation expects the keys to be parsed and converted to node objects prior to using them.

masad-frost: Right, thanks for suggestion

lhchavez (approved):
:tokelau: 

hooray for working CI!

PR #87, opened 2021-03-19
Allow opening channels with the same name if others are closing

masad-frost:
Why
===

The protocol forbids opening a channel with the same name, so when a client tries to do that we throw. However, if the user called close on the old named channel, we should allow it. This should be safe from a protocol perspective since the close request should get in before the next open request. Ideally we `await` closing the channel before trying to open a new one, but it ends up being a tedious task to keep the handle for closing around and waiting for it, both from the client's perspective and caller's perspective.

What changed
============
Only disallow opening a channel with the same name if there are any that are not pending a clousre.

Test plan
=========
I wrote a test to cover this case.

### DIFF
--- a/src/__tests__/index.test.ts
+++ b/src/__tests__/index.test.ts
@@ -628,6 +628,39 @@ test('client rejects opening same channel twice', () => {
   }).toThrow();
 });
 
+test('allows opening channel with the same name while if others are closing', (done) => {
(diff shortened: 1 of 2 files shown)


masad-frost:
E R G O N O M I C S

moudy (approved):
🤸‍♂️

PR #92, opened 2021-05-19
Fix grammar and typos

donno2048:
Why
===

 - "again" is misspelled
 - "queue message" makes no sense
 - prevent **from** being

What changed
============

 - Rewrite "agian" as "again"
 - Add a "s" (queue message **s**)
 - Add this "from"

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ Read about the protocol here http://protodoc.turbio.repl.co
 
 ### Usage and concepts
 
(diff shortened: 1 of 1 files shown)


firefish111:
Ok, no need to be so pedantic

lhchavez (approved):
these changes are great! most of us don't have English as a first language, and we make typos every now and then.

donno2048:
Glad I can help

PR #94, opened 2021-05-24
Relax the criteria to consider the WebSockets implementation as "working"

lhchavez:
Why
===

What changed
============

Previously, it was required to receive a valid api.Command over the WS
connection to deem the WebSockets implementation as "working". When
there is an incident, some of the WS connections are rejected early and
clients will unnecessarily fall back to using the polling
implementation, causing unnecessary churn.

This change makes the criteria to consider the websockets as working to
be just the fact that the `onopen` event was fired.

Test plan
=========

```shell
yarn test
```

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -882,6 +882,11 @@ export class Client<Ctx extends unknown = null> {
     this.ws = ws;
     this.connectionMetadata = connectionMetadata;
 
+    // We'll use this to determine whether or not we should consider the next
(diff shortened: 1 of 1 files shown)


sergeichestakov (approved):
🕸️

masad-frost:
This is a very good change, thank you!

masad-frost:
Are firewall systems upgrade your request and send you back gibberish?

PR #95, opened 2021-05-24
Make the tests a little bit more robust

lhchavez:
Why
===

What changed
============

Previously, `done` could be called multiple times per test (in theory,
at least). Moreover, the `done` callback would be called before the
client was fully torn down. Both of these things would cause unrelated
tests to fail in "impossible" ways.

This change makes the tests always try very hard to call `done` exactly
once in the channel 0 callback, and if it's not possible, do that in a
place that guarantees that the client won't continue doing stuff. Since
now this requires all of the changed tests to clean up after themselves
and in some cases actually hear back from goval, the timeout has been
bumped a bit to account for that.

Also, all the `setUnrecoverableErrorHandler()` calls that were not
handling test logic were removed. Tests will fail _loudly_ due to the
thrown exception.

Test plan
=========

```shell
yarn test
```


sergeichestakov (approved):
Looks good! 🤖

sergeichestakov (commented):
As an aside, we shoulda probably set up CI for this repo. GH action that runs `yarn test` should be pretty easy to do

lhchavez:
> As an aside, we shoulda probably set up CI for this repo. GH action that runs `yarn test` should be pretty easy to do

do you mean something like https://github.com/replit/crosis/runs/2657197905?check_suite_focus=true ? if so, yes, it _was_ easy!

sergeichestakov:
Oh solid! haha I must just be behind the times then

masad-frost:
I just saw this PR.

> This change makes the tests always try very hard to call done exactly once in the channel 0 callback, and if it's not possible, do that in a place that guarantees that the client won't continue doing stuff.

You want to call `done` after your `expects` are done. The `channel 0` callback can happen independently from the places where you call expect. The test will be concluded as successful, even though you didn't assert the things you need to assert.

> Previously, done could be called multiple times per test (in theory, at least).

This is ok, one call is for errors, and the other is for success. If the test fails, it doesn't matter if we call it done elsewhere (though, in theory, the process should stop after the first call).


> Also, all the setUnrecoverableErrorHandler() calls that were not handling test logic were removed. Tests will fail loudly due to the thrown exception.

All the tests expect nothing wrong to happen, so we should keep it. It saved me a few times.


I'm gonna go ahead and undo some of the changes here.

masad-frost:
If you have examples where this caused issues, let me know.

lhchavez:
> I'm gonna go ahead and undo some of the changes here.

as long as tests keep working as reliably as they are right now, go for it.

> If you have examples where this caused issues, let me know.

i don't remember, haha, but the tests were flakiness capital and failed / timed out left and right.

lhchavez:
> This is ok, one call is for errors, and the other is for success. If the test fails, it doesn't matter if we call it done elsewhere (though, in theory, the process should stop after the first call).

i *think* this was the source of some of the problems: the process was not being torn down, so some of the code was still being executed in parallel with another test, and then that caused Jest to be very confused and fail in mysterious ways because it was still executing code from a different test and interfering with the current one.

masad-frost:
Ah sounds good. Lemme run the tests many times and see where the failures are happening. Hopefully it's all network issues

lhchavez:
> Ah sounds good. Lemme run the tests many times and see where the failures are happening. Hopefully it's all network issues

i'm afraid not, haha.

there are more races, which is why some additional guarantees around clean test teardown were added. but if there are other ways of achieving this that can still guarantee test robustness, :+1:.

PR #96, opened 2021-05-24
Allow recycling of connection metadata upon reconnects

lhchavez:
Why
===

What changed
============

Getting new tokens every time we want to reconnect is slightly wasteful,
since most likely the token the clients have already are still valid.
Furthermore, during incidents where goval is unhappy, all the extra
retries will cause cascading failures due to all the attempts to re-get
tokens, which causes extra load to repl-it-web as well as Lore.

This change now keeps the connection metadata across reconnects,
_unless_ goval explicitly tells crosis to fetch a new one. This happens
when goval accepts the connection and immediately closes it with the
Close Code 1008 (Policy Violation).

Test plan
=========

```shell
yarn test
```

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -785,81 +785,85 @@ export class Client<Ctx extends unknown = null> {
     });
     this.channels[0] = chan0;
 
-    if (this.fetchTokenAbortController) {
-      this.onUnrecoverableError(new Error('Expected fetchTokenAbortController to be null'));
+    if (this.connectionMetadata === null) {
+      if (this.fetchTokenAbortController) {
+        this.onUnrecoverableError(new Error('Expected fetchTokenAbortController to be null'));
 
-      return;
-    }
+        return;
+      }
 
-    const abortController = new AbortController();
-    this.fetchTokenAbortController = abortController;
+      const abortController = new AbortController();
+      this.fetchTokenAbortController = abortController;
 
-    let connectionMetadataFetchResult;
-    try {
-      connectionMetadataFetchResult = await this.connectOptions.fetchConnectionMetadata(
-        abortController.signal,
-      );
-    } catch (e) {
-      this.onUnrecoverableError(e);
+      let connectionMetadataFetchResult;
+      try {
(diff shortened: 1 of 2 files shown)


src/client.ts
```diff
@@ -909,13 +918,27 @@ export class Client<Ctx extends unknown = null> {
...
         return;
       }
 
+      if (WebSocketClass === EIOCompat) {
+        if (!didReceiveAnyCommand) {
+          // The polling implementation doesn't convey the Websocket close
+          // event. Let's assume that we need to request a new token.
+          this.connectionMetadata = null;
+        }
+      } else if (Object.prototype.hasOwnProperty.call(event, 'code')) {
```
sergeichestakov, line 934: there should be an easier way to do this. think this should work:
```suggestion
      } else if ('code' in event) {
```

src/client.ts
```diff
@@ -909,13 +918,27 @@ export class Client<Ctx extends unknown = null> {
...
+        const closeEvent = <CloseEvent>event;
+        if (closeEvent.code === 1008) {
```
sergeichestakov, line 936: can we move this up to a const that makes it clear what 1008 means?

src/client.ts
```diff
@@ -909,13 +918,27 @@ export class Client<Ctx extends unknown = null> {
     /**
      * Abrupt socket closures should report failed
      */
-    ws.onclose = () => {
+    ws.onclose = (event: CloseEvent | Event) => {
```
sergeichestakov, line 921: curious is `CloseEvent` not a subset of `Event`? wonder if this is redundant

sergeichestakov (approved):
Nice! Left a couple comments ♻️

src/client.ts
lhchavez: it is (https://developer.mozilla.org/en-US/docs/Web/API/Event), _but_ this can receive both the `CloseEvent` (through the real WebSocket implementation) as well as synthetic `Event`s from the EIO compat layer if we ever fall back to polling.

PR #100, opened 2021-06-30
Make sure to send close requests before opening channels

masad-frost:
Why
===

https://github.com/replit/crosis/pull/87 `(Allow opening channels with the same name if others are closing)` was not implemented correctly. The test there only works because the client is not connected. If the client is already connected and we try to open a channel after calling `close` then try to open the channel again it will lead to a protocol error. This happens because the client will send a new open request before it sends a request to close the current channel, the main struggle is that you actually need the first channel to finish opening so you can use the provided ID to close it. 

What changed
============
- If there are channels with the same name that are in the process of closing, we don't request to open the channel immediately from the `openChannel` call.
- After a channel is done closing, we take the next open request from the top of the queue and open it.

A couple of notes. We will still call the open channel callback, but the channel's state will be `closing`, you can still listen to commands (`channel.onCommand`) as the client can be receiving commands for that channel while it's in the process of closing it. We also do not collapse open/close requests, if the client is open, we will always open the channel then send a close request, I decided to do this from a simplicity standpoint, both from a user perspective and implementation.

N.B There are a cleaner ways of implementing this whole thing, but they require refactoring the client and it's not worth it. 

Test plan
=========
I added 2 tests, let me know if they need commenting, but they should cover all new and old codepaths for opening channel with the same name.

### DIFF
--- a/src/__tests__/index.test.ts
+++ b/src/__tests__/index.test.ts
@@ -618,7 +618,6 @@ test('closing client maintains openChannel requests', (done) => {
 
 test('client rejects opening same channel twice', () => {
   const client = new Client();
-  client.setUnrecoverableErrorHandler(() => {});
 
   const name = Math.random().toString();
   client.openChannel({ name, service: 'exec' }, () => {});
@@ -630,7 +629,6 @@ test('client rejects opening same channel twice', () => {
 
 test('allows opening channel with the same name after closing others and client is disconnected', (done) => {
   const client = new Client();
-  client.setUnrecoverableErrorHandler(done);
 
   const name = Math.random().toString();
 
@@ -665,7 +663,6 @@ test('allows opening channel with the same name after closing others and client 
 
 test('allows opening channel with the same name after others are closing others and client is connected', (done) => {
   const client = new Client();
-  client.setUnrecoverableErrorHandler(done);
 
   client.open(
     {
--- a/src/__tests__/index.test.ts
+++ b/src/__tests__/index.test.ts
@@ -634,12 +634,16 @@ test('allows opening channel with the same name after closing others and client 
 
   const name = Math.random().toString();
 
-  const close = client.openChannel({ name, service: 'exec' }, () => {});
+  let calledFirstWithError = false;
+  const close = client.openChannel({ name, service: 'exec' }, ({ error }) => {
+    calledFirstWithError = Boolean(error);
+  });
 
   close();
   // open same name synchronously
   client.openChannel({ name, service: 'exec' }, ({ channel }) => {
     expect(channel).toBeTruthy();
+    expect(calledFirstWithError).toBeTruthy();
     client.close();
 
     done();
--- a/src/__tests__/index.test.ts
+++ b/src/__tests__/index.test.ts
@@ -734,6 +734,71 @@ test('allows opening channel with the same name after others are closing others 
   );
 });
 
+test('opens multiple anonymous channels while client is connected', (done) => {
+  const client = new Client();
+
+  let didDone = false;
+  const doneOnce = (e?: Error) => {
+    if (didDone) {
+      return;
+    }
+
+    didDone = true;
+    client.close();
+    done(e);
+  };
+
+  client.open(
+    {
+      fetchConnectionMetadata: () =>
+        Promise.resolve({
+          ...genConnectionMetadata(),
+          error: null,
+        }),
+      WebSocketClass: WebSocket,
+      context: null,
+    },
+    ({ channel: chan0 }) => {
+      expect(chan0).toBeTruthy();
+
+      let firstOpened = false;
+      let secondOpened = false;
+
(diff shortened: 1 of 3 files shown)


src/__tests__/index.test.ts
```diff
@@ -653,8 +655,77 @@ test('allows opening channel with the same name while if others are closing', (d
...
+      }
+
+      const name = Math.random().toString();
+
+      let firstChannel: Channel;
+      const close = client.openChannel({ name, service: 'exec' }, ({ channel }) => {
+        expect(channel).toBeTruthy();
+
+        if (!channel) {
+          throw new Error('apease typescript');
```
lhchavez, line 690: https://github.com/microsoft/TypeScript/issues/198 :(

src/__tests__/index.test.ts
```diff
@@ -627,20 +628,21 @@ test('client rejects opening same channel twice', () => {
   }).toThrow();
 });
 
-test('allows opening channel with the same name while if others are closing', (done) => {
+test('allows opening channel with the same name after closing others and client is disconnected', (done) => {
   const client = new Client();
+  client.setUnrecoverableErrorHandler(done);
```
lhchavez, line 633: quick hack to still achieve "one `done()` per test" (and as a bonus tells the test runner that there's something wrong _somewhere else_):

```typescript
function oneShot(f: (...args: any) => void): (...: any) => void {
  let called = false;
  return (...args: any) => {
    if (called) {
      console.error("called function more than once. this is likely a bug in another test.");
      return;
    }
    called = true;
    f(...args);
  };
}
```

and then wrap the original `done` with the oneshot version. that way tests cannot contaminate other tests (my worry) _and_ the unrecoverable error handler can still be there (your worry). everybody wins!

src/client.ts
```diff
@@ -1323,6 +1354,12 @@ export class Client<Ctx extends unknown = null> {
           willReconnect: willChannelReconnect,
         });
       }
+
+      if (closeRequested || channelRequest.closeRequested) {
```
lhchavez, line 1358: since `closeRequested` is only ever assigned to `true`, the combo `closeRequested === true && channelRequest.closeRequested === false` is unreachable, meaning that `channelRequest.closeRequested` is sufficient.

src/client.ts
```diff
@@ -363,23 +364,36 @@ export class Client<Ctx extends unknown = null> {
 
     if (this.connectionState === ConnectionState.CONNECTED) {
       // We're connected, open channel. Otherwise we'll open the channel once we connect
-      this.requestOpenChannel(channelRequest);
+      if (!sameNameChanRequests.length) {
```
lhchavez, line 367: maybe collapse L365-L367 in a single condition to make the comments make more sense.

also, does this also need the `options.name &&`? or is there something about anon channel requests that make this work correctly?

src/client.ts
```diff
@@ -612,6 +629,20 @@ export class Client<Ctx extends unknown = null> {
...
+
+    const nextRequest = this.channelRequests.find(
+      (cr) => cr.options.name === channelRequest.options.name,
+    );
+
+    if (!nextRequest) {
+      return;
+    }
+
+    this.requestOpenChannel(nextRequest);
```
lhchavez, line 645: can this block have a comment? i understand what it's doing, but not _why_ we are opening the channel here. is it because we want to wait until the previous close request finished before issuing a new one?

lhchavez (approved):
:mailbox_closed: 

to the best of my abilities, this looks good! thanks for fixing this weird race.

src/client.ts
masad-frost: Big brain

src/client.ts
masad-frost: Exactly. In theory we can send the open request as soon as we send the `closeChan` request, but I thought I'd just keep it simple and do everything sequentially.

Will add a comment

src/client.ts
masad-frost: Oof, yes needs `options.name &&`. Gonna try and make a test that fails with this in place.

src/__tests__/index.test.ts
masad-frost: Good idea. I'll remove this completely for now and revisit with a separate PR

src/client.ts
masad-frost: Oh wait, on line 1315 we set it to false. `closeRequested` may be true from earlier, `clanupCb` may not call close on the channel and so `channelRequest.closeRequested` maybe false here, `closeRequested === true && channelRequest.closeRequested === false` can happen. 

There's a world where we don't have to think about this and it happens in one place!

src/__tests__/index.test.ts
masad-frost: Oh, that would be cool! cc @allawesome497 we talked about this the other day.

src/client.ts
masad-frost: Very good catch btw, this could've been disastrous!

PR #101, opened 2021-07-01
Unset connection metadata when close is intentional

masad-frost:
Why
===
We had this guarantee that we will get new connection metadata after each disconnect, intentional or not. Now we always keep it cached, unless there's a specific websocket error.

I _think_  some parts of the website rely on the old behavior for intentional closes, the hint is provided by this code here https://github.com/replit/repl-it-web/blob/master/app/services/container/index.ts#L358-L361. More specifically, I think the anonymous language page uses this.

What changed
============
As the title said.

Test plan
=========
I should probably add a test, brb

### DIFF
--- a/src/__tests__/index.test.ts
+++ b/src/__tests__/index.test.ts
@@ -641,7 +641,9 @@ test('closing client maintains openChannel requests', (done) => {
             WebSocketClass: WebSocket,
             context: null,
           },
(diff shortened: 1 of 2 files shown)


lhchavez (approved):
:closed_book: 

this is super reasonable. whatever test is written should be good!

PR #103, opened 2021-07-02
Add a way to register more than one debug callback

frantic:
Why
===

It's useful to have more than one callback at a time. For example, one for Sentry breadcrumbs reporting, the other one for debug panel or dumping logs to a text file in test environment.

What changed
============

Added new `addDebugFunc` method that returns a cleanup function. The callbacks are stored in a plan array for simplicity and portability — didn't want to depend on `EventEmitter` or `Map`.

The old `setDebugFunc` API still works.

Test plan
=========

👀

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -125,11 +125,19 @@ export class Client<Ctx extends unknown = null> {
   };
 
   /**
-   * Called for breadcrumbs and other debug reasons
+   * Called for breadcrumbs and other debug reasons. Use addDebugFunc
+   * instead
    *
    * @hidden
    */
-  private debug: (log: DebugLog) => void;
+  private legacyDebugFunc: undefined | ((log: DebugLog) => void);
+
+  /**
+   * Listeners to be called for breadcrumbs and other debug reasons
+   *
+   * @hidden
+   */
+  private debugFuncs: Array<(log: DebugLog) => void>;
 
   /**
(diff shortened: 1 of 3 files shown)


src/client.ts
```diff
@@ -726,9 +734,32 @@ export class Client<Ctx extends unknown = null> {
     return chan;
   };
 
+  /** Sets a logging/debugging function */
```
lhchavez, line 737:
```suggestion
  /**
   * Calls all the registered logging/debugging functions
   *
   * @hidden
   */
```

src/client.ts
```diff
@@ -726,9 +734,32 @@ export class Client<Ctx extends unknown = null> {
...
+  private debug = (log: DebugLog): void => {
+    if (this.legacyDebugFunc) {
+      this.legacyDebugFunc(log);
+    }
+    this.debugFuncs.forEach((func) => func(log));
+  };
+
   /** Sets a logging/debugging function */
```
lhchavez, line 745:
```suggestion
  /**
   * Sets a logging/debugging function
   *
   * @deprecated use addDebugFunc instead
   */
```

lhchavez (approved):
(although maybe also wait for @masad-frost's opinion

masad-frost (approved):
Works for me!

PR #105, opened 2021-07-10
Reland "Allow recycling of connection metadata upon reconnects (#96)"

lhchavez:
Why
===

Getting new tokens every time we want to reconnect is slightly wasteful,
since most likely the token the clients have already are still valid.
Furthermore, during incidents where goval is unhappy, all the extra
retries will cause cascading failures due to all the attempts to re-get
tokens, which causes extra load to repl-it-web as well as Lore.

This change is better reviewed with [whitespace disabled :dark_sunglasses:](https://github.com/replit/crosis/pull/105/files?diff=split&w=1).

What changed
============

This is a mostly as-is reland of #96 (by cherry-picking that change, instead of reverting the revert).

This change now keeps the connection metadata across reconnects,
_unless_ goval explicitly tells crosis to fetch a new one. This happens
when goval accepts the connection and immediately closes it with the
Close Code 1008 (Policy Violation).

This time around, the `ConnectOptions` now receives a
`reuseConnectionMetadata` field, and will reuse tokens only when it's
set to `true`. This allows a flagged deployment.

Test plan
=========

```shell
yarn test
```

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -259,6 +259,7 @@ export class Client<Ctx extends unknown = null> {
 
     this.connectOptions = {
       timeout: 10000,
+      reuseConnectionMetadata: false,
       ...options,
     };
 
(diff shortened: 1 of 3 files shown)


src/client.ts
```diff
@@ -259,6 +259,7 @@ export class Client<Ctx extends unknown = null> {
 
     this.connectOptions = {
       timeout: 10000,
+      reuseConnectionMetadata: false,
```
jeremypress, line 262: don't we need to check the options value? We can still default to false for safety

jeremypress, line 262: ah I misread this, nice override below

jeremypress (approved):

PR #109, opened 2021-09-09
Allow polling host to be overridden

miselin:
Why
===

We're moving gp-v2 to gp-v2.replit.com. Being able to override the polling host in the client will make that migration easier.

What changed
============

Added `pollingHost` to `ConnectOptions`. We'll be able to override in client code and roll back changes without having to roll back this PR.

Test plan
=========

Block eval.global.replit.com via `/etc/hosts`, see requests still go to gp-v2.herokuapp.com.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -954,7 +954,7 @@ export class Client<Ctx extends unknown = null> {
     const WebSocketClass = isPolling
       ? EIOCompat
       : getWebSocketClass(this.connectOptions.WebSocketClass);
-    const connStr = getConnectionStr(this.connectionMetadata, isPolling);
+    const connStr = getConnectionStr(this.connectionMetadata, isPolling, this.connectOptions.pollingHost);
     const ws = new WebSocketClass(connStr);
 
     ws.binaryType = 'arraybuffer';
--- a/src/util/helpers.ts
+++ b/src/util/helpers.ts
@@ -57,7 +57,7 @@ export function getWebSocketClass(
 export function getConnectionStr(
   connectionMetadata: GovalMetadata,
   isPolling: boolean,
-  pollingHost: string | null,
+  pollingHost?: string,
 ): string {
   const gurl = urllib.parse(connectionMetadata.gurl);
   if (isPolling) {
--- a/src/util/helpers.ts
+++ b/src/util/helpers.ts
@@ -61,7 +61,7 @@ export function getConnectionStr(
 ): string {
   const gurl = urllib.parse(connectionMetadata.gurl);
   if (isPolling) {
-    const host = pollingHost || 'gp-v2.herokuapp.com';
+    const host = pollingHost ?? 'gp-v2.herokuapp.com';
     gurl.hostname = host;
     gurl.host = host;
     gurl.pathname = `/wsv2/${connectionMetadata.token}/${encodeURIComponent(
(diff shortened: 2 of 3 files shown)


src/util/helpers.ts
```diff
@@ -57,8 +57,8 @@ export function getWebSocketClass(
 export function getConnectionStr(connectionMetadata: GovalMetadata, isPolling: boolean): string {
   const gurl = urllib.parse(connectionMetadata.gurl);
   if (isPolling) {
-    gurl.hostname = 'gp-v2.herokuapp.com';
-    gurl.host = 'gp-v2.herokuapp.com';
+    gurl.hostname = 'gp-v2.replit.com';
+    gurl.host = 'gp-v2.replit.com';
```
lhchavez, line 61: can these be passed as parameters in the options structure? that way we can roll it back and forward all the times we want / need (ideally times = 1).

miselin: yep! Made that change, it'll make migrating much easier.

src/client.ts
```diff
@@ -954,7 +954,7 @@ export class Client<Ctx extends unknown = null> {
     const WebSocketClass = isPolling
       ? EIOCompat
       : getWebSocketClass(this.connectOptions.WebSocketClass);
-    const connStr = getConnectionStr(this.connectionMetadata, isPolling);
+    const connStr = getConnectionStr(this.connectionMetadata, isPolling, this.connectOptions.pollingHost);
```
lhchavez, line 957: this probably needs

```diff
diff --git a/src/types.ts b/src/types.ts
index 2cf95fb..83ba169 100644
--- a/src/types.ts
+++ b/src/types.ts
@@ -43,6 +43,7 @@ export interface ConnectOptions<Ctx> {
   WebSocketClass?: typeof WebSocket;
   context: Ctx;
   reuseConnectionMetadata: boolean;
+  pollingHost?: string;
 }

 export interface UrlOptions {
```

src/util/helpers.ts
```diff
@@ -54,11 +54,16 @@ export function getWebSocketClass(
 /**
  * Given connection metadata, creates a websocket connection string
  */
-export function getConnectionStr(connectionMetadata: GovalMetadata, isPolling: boolean): string {
+export function getConnectionStr(
+  connectionMetadata: GovalMetadata,
+  isPolling: boolean,
+  pollingHost: string | null,
```
lhchavez, line 60:
```suggestion
  pollingHost?: string,
```

for type congruence with the above suggestion.

src/util/helpers.ts
```diff
@@ -54,11 +54,16 @@ export function getWebSocketClass(
...
+): string {
   const gurl = urllib.parse(connectionMetadata.gurl);
   if (isPolling) {
-    gurl.hostname = 'gp-v2.herokuapp.com';
-    gurl.host = 'gp-v2.herokuapp.com';
+    const host = pollingHost || 'gp-v2.herokuapp.com';
```
lhchavez, line 64:
```suggestion
    const host = pollingHost ?? 'gp-v2.herokuapp.com';
```

the [nulling coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator) is a bit better for missing (rather than falsy) stuff.

lhchavez (approved):

src/client.ts
miselin: whoops, left this out of the commit

PR #114, opened 2022-01-08
Stop giving a default `pollingHost` if unset

lhchavez:
Why
===

Previously, we would provide a default of `gp-v2.replit.com` when
falling back to polling. But some folks might not _want_ to fallback to
polling.

What changed
============

This change now requires clients to explicitly set a `pollingHost` if
they want to fallback to *something*.

Test plan
=========

Added a test!


miselin (approved):
🛑

PR #117, opened 2022-03-11
Expose channel name and service name in debug logs

cbrewster:
Why
===

We want to expose channel name and service name in the logs so we can add improved logging filters in the admin panel.

What changed
============

- Store channel name and service name on the `Channel`
- Include channel name and service name in debug log messages

Test plan
=========

- Shrug

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -512,6 +512,8 @@ export class Client<Ctx extends unknown = null> {
 
       const channel = new Channel({
         id,
+        name: options.name,
+        service,
         onUnrecoverableError: this.onUnrecoverableError,
         send: this.send,
       });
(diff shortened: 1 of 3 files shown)


src/client.ts
```diff
@@ -950,14 +954,14 @@ export class Client<Ctx extends unknown = null> {
       });
     }
 
-    const isPolling =
-        websocketFailureCount >= 3 && this.connectOptions.pollingHost;
+    const isPolling = websocketFailureCount >= 3 && this.connectOptions.pollingHost;
```
cbrewster, line 957: sorry prettier decide to format some things, looks better imo

airportyh (approved):

PR #119, opened 2022-05-09
Provide channel service / name on the crosis logs

lhchavez:
Why
===

Previously one had to pay attention and manually correlate what channels
correspond to what services when inspecting the crosis logs. That's a
bit too hard for humans, and the machines can do that for us.

What changed
============

This change adds the channel service (if available) and name (if
available) for all crosis logs.

Test plan
=========

image

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -512,6 +512,8 @@ export class Client<Ctx extends unknown = null> {
 
       const channel = new Channel({
         id,
+        name: channelRequest.options.name,
+        service,
         onUnrecoverableError: this.onUnrecoverableError,
         send: this.send,
       });
@@ -856,6 +858,7 @@ export class Client<Ctx extends unknown = null> {
 
     const chan0 = new Channel({
       id: 0,
+      name: 'chan0',
       onUnrecoverableError: this.onUnrecoverableError,
       send: this.send,
     });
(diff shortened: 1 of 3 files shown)


src/client.ts
```diff
@@ -950,14 +953,14 @@ export class Client<Ctx extends unknown = null> {
...
     const WebSocketClass = isPolling
       ? EIOCompat
       : getWebSocketClass(this.connectOptions.WebSocketClass);
-    const connStr = getConnectionStr(this.connectionMetadata,
-                                     isPolling ? this.connectOptions.pollingHost
-                                               : undefined);
+    const connStr = getConnectionStr(
+      this.connectionMetadata,
+      isPolling ? this.connectOptions.pollingHost : undefined,
+    );
```
lhchavez, line 963: this was all `prettier`'s fault.

cbrewster (approved):
Haha never hit merge on mine https://github.com/replit/crosis/pull/117

src/types.ts
```diff
@@ -77,6 +77,10 @@ export type DebugLog =
       type: 'log';
       log: {
         direction: 'in' | 'out';
+        channel: {
+          name?: string;
```
masad-frost, line 81: Might as well throw the ID in there as well.

masad-frost: A little redundant since cmd should have it, but I think it makes sense

masad-frost (approved):

src/types.ts
lhchavez: might as well!

PR #120, opened 2022-05-11
Remove retryTimeoutId after closing

masad-frost:
Why
===
We see `unexpected existing retryTimeoutId` in our error logs. 

What changed
============
I cleared `retryTimeoutId` inside `handleClose` when we clear the timeout.

Test plan
=========
I'm gonna add a test in a follow up PR that handles this case and more reconnect/retry cases, but here's how to reproduce this current issue:
1. Open a client
2. Get it in a state where it's reconnecting
3. Just before it reconnects call `close` on the client
4. Re-open the client
5. Get it in a state where it's reconnecting
6. ERROR

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -950,14 +950,14 @@ export class Client<Ctx extends unknown = null> {
       });
     }
 
-    const isPolling =
-        websocketFailureCount >= 3 && this.connectOptions.pollingHost;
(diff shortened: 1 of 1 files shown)


sergeichestakov (approved):
Thanks! good catch ⏲️

PR #121, opened 2022-05-12
Fm rework retry

masad-frost:
WIP just putting it up for visibility. The gist of it is to delegate some of the retry logic to the user

Why
===
We need to delegate more of retry logic to clients and applications. For example, change retry rate based on external factors, or handle retries completely in UI.

What changed
============
- Delegate retries to caller
- Emit websocket error codes and other reconnect reasons
- Allow for aborting

Test plan
=========
WIP


masad-frost:
we shipped a less complete version of this here https://github.com/replit/crosis/pull/142, it's good enough for now.

PR #127, opened 2022-05-13
Fix timeouts lingering after closing

masad-frost:
Why
===
If we have a pending connection with a timeout option and we close in the middle of it connecting, the timeout lingers and leads to undefined behavior, from my testing it seems like it bottoms out here https://github.com/replit/crosis/blob/711c2bd7079b426857d4a41a0a6959e2ebac89f0/src/client.ts#L1233-L1238 which seems innocent, but the problem is that it calls https://github.com/replit/crosis/blob/711c2bd7079b426857d4a41a0a6959e2ebac89f0/src/client.ts#L1198-L1200 before getting there, leading to a spooky undefined state.

<sup><sup><sup><sup><sup><sup><sup>This whole library feels like it should much much simpler! Coroutines? Reactivity? State machine?</sup></sup></sup></sup></sup></sup></sup>

What changed
============
- Added check inside `onFailed` to make sure we don't have multiple connections tries going on at once. It checks referential equality for things unique to a connect try.
- Introduce `connectTimeoutId` property on the class, use it during closing to clear up the timeout (this is the fix)
- Moved abort controller aborting to a common cleanup place
- Added a test
- Made `getClient` in tests require `done` so that it passes it as an unrecoverable error handler and fail the test
- Made `WebsocketThatNeverConnects` in testing follow `close` semantics of real websockets

Test plan
=========
Other than adding the test, the breadcrumbs prove that we don't have any lingering timeouts.

### DIFF
--- a/src/__tests__/index.test.ts
+++ b/src/__tests__/index.test.ts
@@ -1238,7 +1238,7 @@ test('does not fallback to polling if host is unset', (done) => {
   );
 }, 40000);
 
-test.only('cancels connection timeout when closing', (done) => {
+test('cancels connection timeout when closing', (done) => {
   const client = getClient(done);
 
   const WebsocketThatNeverConnects = getWebsocketClassThatNeverConnects();
--- a/src/__tests__/index.test.ts
+++ b/src/__tests__/index.test.ts
@@ -64,6 +64,13 @@ function wrapWithDone<Args extends Array<any>, Ret>(
 function getWebsocketClassThatNeverConnects() {
   class _WebsocketThatNeverConnects {
     static OPEN = 1;
+    onclose?: (closeEvent: CloseEvent) => void;
+
+    send = () => {};
+
+    close = (code = 1000, reason?: string) => {
+      setTimeout(() => this.onclose?.(createCloseEvent({ code, reason })));
+    };
   }
 
   const WebsocketThatNeverConnects = _WebsocketThatNeverConnects as typeof WebSocket;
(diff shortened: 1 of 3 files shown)


src/__tests__/index.test.ts
```diff
@@ -61,8 +63,30 @@ function wrapWithDone<Args extends Array<any>, Ret>(
...
+  // Make sure internals think it's a websocket
+  expect(() => getWebSocketClass(WebsocketThatNeverConnects)).not.toThrow();
+  expect(getWebSocketClass(WebsocketThatNeverConnects)).toEqual(WebsocketThatNeverConnects);
+
+  return WebsocketThatNeverConnects;
+}
+
 test('client connect', (done) => {
-  const client = getClient<{ username: string }>();
+  const client = getClient<{ username: string }>(done);
```
lhchavez, line 89: FYI the last time we had multiple `done` calls, it confused jest significantly and caused a non-trivial amount of flake and lack of hermeticity. we may need to wrap this in an at-most-once wrapper to prevent that from happening again: https://github.com/replit/crosis/pull/95 .

either that or wire up an `AbortController` so that we can tear things down when we know things are going to fail (and then test that it gracefully stops things no matter where in the processing it is).

then again, that whole ordeal may only have happened when there _were_ failures / flakes, since when tests pass, `done` is only called once.

lhchavez (approved):

src/__tests__/index.test.ts
lhchavez: oh nice! one fewer thing to worry about then!

PR #128, opened 2022-05-23
Deny access to unfirewalled repls from firewall mode

lincoln-replit:
Why
===

This is a companion to https://github.com/replit/goval/pull/4964

The above PR adds the FirewallDenied message to @replit/protocol, which is sent when a user tries to connect to an unfirewalled repl from Firewall Mode.

Why not do this all client side? Why does goval need to be involved at all? Two reasons:
1. Whether a repl is currently running in firewall mode or not is not known to web; only goval
2. Security reasons; if firewalledreplit.com was capable of minting an unfirewalled token that eval.global.replit.com could accept, we wouldn't be able to guarantee that all repls opened from firewall mode cannot access the broader internet

What changed
============

Check for the FirewallDenied command from the protocol, and throw an unrecoverable error if so.

Also bump package versions, will publish after PR is accepted

Test plan
=========

yarn link to repl-it-web, run with https://github.com/replit/goval/pull/4964, confirm that firewall mode cannot connect to unfirewalled repls (see goval PR for more details)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@replit/crosis",
-  "version": "v8.0.1",
+  "version": "v8.0.0",
   "description": "Goval connection and channel manager",
   "files": [
     "/dist"


src/client.ts
```diff
@@ -1125,6 +1125,12 @@ export class Client<Ctx = null> {
       // we will reset the timeout
       resetTimeout();
 
+      if (cmd.FirewallDenied === true) {
```
lhchavez, line 1128: is it worth letting this command go all the way to repl-it-web? that way that can be handled however we want there (plus adding fewer edge cases in this repository).

lhchavez: (also, i18n later)

lincoln-replit: I'm open to it. I think to make it work we would need to add another callback, something like `this.onFirewallDenied()`

lincoln-replit: On second thought, I don't think that would actually reduce complexity in this library or make it easier to use -- unless I'm misunderstanding, did you have something specific in mind?

lincoln-replit: @masad-frost alluded to a rewrite of this library the other day, perhaps it can be considered there. This approach as-is cargo cults the existing style.

masad-frost: I don't think we're gonna rewrite this library until we rewrite our whole protocol (unclear when that will happen).

> On second thought, I don't think that would actually reduce complexity in this library or make it easier to use -- unless I'm misunderstanding, did you have something specific in mind?

I don't think Luis is referring to complexity, just more about explicitness. So like maybe we want to show a specific modal instead of a generic BSOD.

Going to unblock for now, but could be a good follow up.

I need to add a way to capture messages before we're "connected" (like bootstatus), instead of having on-off listeners like `onBootStatus` and `onFirewallDenied` maybe I'll add `onConnectingMessage` or something like that. Will keep you looped in.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@replit/crosis",
-  "version": "v8.0.0",
+  "version": "v8.0.1",
```
masad-frost, line 3: Can you undo this? We'll bump the version/git tag in a separate commit.

masad-frost (approved):
If we can add a test, it'd be great. It'll take me a couple of minutes to add, so I can follow up, but i need you to tell me what needs to go into the signed token here https://github.com/replit/crosis/blob/main/debug/genConnectionMetadata.js to make that work

package.json
lincoln-replit: Done

lincoln-replit:
@masad-frost In the resourceLimits object after [this line](https://github.com/replit/crosis/blob/main/debug/genConnectionMetadata.js#L48), add `restrictNetwork: true`.

See the tests for https://github.com/replit/goval/pull/4964, but basically:
- Initial connect with restrictNetwork -> all future connections with or without restrictNetwork are allowed
- Initial connect without restrictNetwork -> future connections that use restrictNetwork won't work

src/client.ts
lhchavez: yup, it was a request about explicitness. doing a generic BSOD for now sounds reasonable, just wasn't sure if it was quick and easy enough to push that knowledge directly to repl-it-web in a way that didn't make sense to modify this package at all.

seems like it's not the case, so onwards!

PR #129, opened 2022-06-01
Rename addDebugFunc to onDebugLog

masad-frost:
Why
===

Matches our existing event handler naming

What changed
============
- Rename `addDebugFunc` to `onDebugLog`
- Removed deprecated `setDebugFunc`

Test plan
=========
We have tests in place that use this, they should pass


sergeichestakov (approved):
✏️

PR #130, opened 2022-06-01
Add a way to listen to boot status messages

masad-frost:
Why
===

There's no way for the user to read control channel messages before we connect and user gets an instance of channel 0. We'll add an explicit listener for boot status messages.

https://app.asana.com/0/1202342654232057/1202342654232071

What changed
============
- Add `onBootStatus` event listener
- call when we get a bootstatus message during connection

Test plan
=========
Added a test


sergeichestakov (approved):
👢

PR #132, opened 2022-06-29
Refactor the firewall denied error condition and add a test

lincoln-replit:
Why
===

Discussed in https://github.com/replit/crosis/pull/128

In order to present a more friendly interface to the user when this happens, so that firewalledreplit.com is not perceived as buggy, let's give ourselves an explicit function to call

Also, add a test so we know it works

What changed
============

- Instead of throwing an error, call a function (still throws an error by default)
- Small test refactor so we can put the new test in its own file

Test plan
=========

Test should be sufficient


lincoln-replit:
Fixed goval, this works now!

masad-frost (approved):
Best javascript slinger in the south. Thanks for following up.

debug/genConnectionMetadata.d.ts
```diff
@@ -1,5 +1,17 @@
-export default function genConnectionMetadata(): {
+type Repl = {
+  id: string;
+  languag: string;
```
xyc, line 3: typo?
```suggestion
  language: string;
```

debug/genConnectionMetadata.js
```diff
@@ -20,9 +20,22 @@ ${keyData.toString('base64')}
...
+  if (repl == null) {
+    repl = {
+      id: `testing-crosis-${Math.random().toString(36).split('.')[1]}`,
+      language: 'bash',
+      slug: Math.random().toString(36).slice(2),
+      user: 'crosistest',
+      bucket: 'test-replit-repls',
+    };
+  }
+
```
xyc, line 38: nit: slightly easier way of writing
```suggestion
  let repl = options?.repl || {
    id: `testing-crosis-${Math.random().toString(36).split('.')[1]}`,
    language: 'bash',
    slug: Math.random().toString(36).slice(2),
    user: 'crosistest',
    bucket: 'test-replit-repls',
  };
```

src/__testutils__/done.ts
```diff
@@ -0,0 +1,38 @@
+/* eslint-env jest */
+
+/**
+ * A simple helper to allow us to do assertions inside callbacks.
+ * If `done` is not called after an assertion fails inside a callback
+ * the test will continue until it times out. This function makes sure
+ * done is called and we fail fast with correct stack traces.
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function wrapWithDone<Args extends Array<any>, Ret>(
```
xyc, line 10: nice refactor

PR #135, opened 2022-07-05
Bump crosis version to 9.0.1

lincoln-replit:
Why
===

I don't know if this change would be considered breaking or not, but the main change is that instead of emitting an unrecoverable error, firewallDenied will call a dedicated callback function that can be set by the client.

What changed
============

9.0.0 -> 9.0.1

Test plan
=========

Tests pass


masad-frost (approved):
ty

PR #137, opened 2022-07-11
Handle redirect message

brenoafb:
Why
===

We're designing a new mechanism that will allow us greater freedom in placing Repls.
This will require support for a redirect protocol message. This message is passed
alongside an address to which the Crosis client should redirect.
This redirection is performed transparently by the client implementation, thus no further
action is needed.

Here is how things should typically go
```
+-----+              +---------+    +-----+
| ws1 |              | crosis  |    | ws2 |
+-----+              +---------+    +-----+
   |                      |            |
   |              connect |            |
   |<---------------------|            |
   |                      |            |
   | redirect to ws2      |            |
   |--------------------->|            |
   |       -------------\ |            |
   |       | disconnect |-|            |
   |       |------------| |            |
   |                      |            |
   |                      | connect    |
   |                      |----------->|
   |                      |            |
```

We also do special handling in case the client disconnects from the server pointed to by a previous redirect.
In this case, we fall back to the server who sent the first redirect.

```
                       +-----+              +---------+                         +-----+
                       | ws1 |              | crosis  |                         | ws2 |
                       +-----+              +---------+                         +-----+
                          |                      |                                 |
                          |              connect |                                 |
                          |<---------------------|                                 |
                          |                      |                                 |
                          | redirect to ws2      |                                 |
                          |--------------------->|                                 |
                          |       -------------\ |                                 |
                          |       | disconnect |-|                                 |
                          |       |------------| |                                 |
                          |                      |                                 |
                          |                      | connect                         |
                          |                      |-------------------------------->|
                          |                      |                                 | -----------------------------------\
                          |                      |                                 |-| disconnect, something went wrong |
                          |                      |                                 | |----------------------------------|
                          |                      | -----------------------------\  |
                          |                      |-| reconnecting to ws1        |  |
                          |                      | | since it sent the redirect |  |
                          |                      | | message to ws2             |  |
                          |              connect | |----------------------------|  |
                          |<---------------------|                                 |
------------------------\ |                      |                                 |
| figure out what to do |-|                      |                                 |
|-----------------------| |                      |                                 |
                          |                      |                                 |
```

What changed
============

- Add a listener to redirect messages in channel 0
- Revert Parcel to v 1.12.3 (v2 doesn't support current dev setup)

Test plan
=========

- `yarn test`

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,8 @@
     "eslint-plugin-prettier": "4.0.0",
     "express": "^4.18.1",
     "jest": "^28.1.0",
+    "jest-websocket-mock": "^2.3.0",
+    "mock-socket": "^9.1.5",
     "parcel": "1.12.3",
     "prettier": "^2.6.2",
     "ts-jest": "^28.0.2",
--- a/package.json
+++ b/package.json
@@ -22,7 +22,8 @@
   "license": "MIT",
   "dependencies": {
     "abortcontroller-polyfill": "^1.7.3",
-    "engine.io-client": "^3.4.0"
+    "engine.io-client": "^3.4.0",
+    "jest-websocket-mock": "^2.3.0"
   },
   "devDependencies": {
     "@replit/protocol": ">=0.2.48",
--- a/package.json
+++ b/package.json
@@ -23,7 +23,8 @@
   "dependencies": {
     "abortcontroller-polyfill": "^1.7.3",
     "engine.io-client": "^3.4.0",
-    "jest-websocket-mock": "^2.3.0"
+    "jest-websocket-mock": "^2.3.0",
+    "mock-socket": "^9.1.5"
   },
   "devDependencies": {
     "@replit/protocol": ">=0.2.48",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@replit/crosis",
-  "version": "9.0.1",
+  "version": "9.0.2",
   "description": "Goval connection and channel manager",
   "files": [
     "/dist"
@@ -22,9 +22,7 @@
   "license": "MIT",
   "dependencies": {
     "abortcontroller-polyfill": "^1.7.3",
-    "engine.io-client": "^3.4.0",
-    "jest-websocket-mock": "^2.3.0",
-    "mock-socket": "^9.1.5"
+    "engine.io-client": "^3.4.0"
   },
   "devDependencies": {
     "@replit/protocol": ">=0.2.48",
@@ -39,7 +37,7 @@
     "eslint-plugin-prettier": "4.0.0",
     "express": "^4.18.1",
     "jest": "^28.1.0",
-    "parcel": "1.12.3",
+    "parcel": "^2.5.0",
     "prettier": "^2.6.2",
     "ts-jest": "^28.0.2",
     "typedoc": "^0.22.15",
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "eslint-plugin-prettier": "4.0.0",
     "express": "^4.18.1",
     "jest": "^28.1.0",
-    "parcel": "^2.5.0",
+    "parcel": "1.12.3",
     "prettier": "^2.6.2",
     "ts-jest": "^28.0.2",
     "typedoc": "^0.22.15",
--- a/src/__tests__/redirect.test.ts
+++ b/src/__tests__/redirect.test.ts
@@ -20,19 +20,7 @@ afterAll(() => {
   });
 });
 describe('websocket', () => {
-
-  test("hello example", async () => {
-    const server = new WS("ws://localhost:1234");
-    const client = new WebSocket("ws://localhost:1234");
-
-    await server.connected;
-    client.send("hello");
-    await expect(server).toReceiveMessage("hello");
-    expect(server).toHaveReceivedMessages(["hello"]);
-    server.close();
-  });
-
-  test('websocket mock', async () => {
+  test('redirect message', async () => {
     const ctx = { username: 'zyzz' };
     // const client = getClient<{ username: string }>(done);
     const client = new Client<{ username: string }>();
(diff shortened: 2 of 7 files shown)


src/client.ts
```diff
@@ -1648,4 +1660,35 @@ export class Client<Ctx = null> {
...
+        fetchConnectionMetadata: () => Promise.resolve(fetchConnectionMetadataResult),
+        WebSocketClass: WebSocket,
+        context: null,
+      },
+      ({ channel, error }) => {
+        console.log('channel ', { channel, error });
+
+        return (reason) => {
+          console.log('channe close after redirect, reason: ', reason);
+        };
```
brenoafb, line 1690: also don't really know if there's anything that needs to be done in these callbacks

src/client.ts
```diff
@@ -1648,4 +1677,50 @@ export class Client<Ctx = null> {
 
     throw e;
   };
+
+  private handleRedirect = (url: string) => {
+    this.debug({
+      type: 'breadcrumb',
```
xyc, line 1683: You might need to add types for the breadcrumb, we recently enforced typing on this: https://github.com/replit/crosis/pull/139

xyc: Update branch and you should see a TypeScript error

src/__tests__/redirect.test.ts
```diff
@@ -0,0 +1,173 @@
...
+import { WebSocket } from 'mock-socket';
+
+// eslint-disable-next-line
+const genConnectionMetadata = require('../../debug/genConnectionMetadata');
+
+// eslint-disable-next-line
+jest.setTimeout(1000);
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const testingClients: Array<Client<any>> = [];
```
xyc, line 15: let's just use `Array<Client{ username: string }>` for the type here?

src/__tests__/redirect.test.ts
```diff
@@ -0,0 +1,173 @@
...
+
+const sendFromServer = (cmd: api.Command, ws: WS) => {
+  const cmdBuf = api.Command.encode(cmd).finish();
+  const buffer = cmdBuf.buffer.slice(cmdBuf.byteOffset, cmdBuf.byteOffset + cmdBuf.length);
+  ws.send(buffer);
+};
+
+const genPortNumber = () => {
+  return Math.floor(Math.random() * 100000);
```
xyc, line 24: If this returns the same port, would `jest-websocket-mock` handle well? (might be rare but in a CI context it might be run a lot, and could get very confusing when it happens)

src/__tests__/redirect.test.ts
```diff
@@ -0,0 +1,173 @@
...
+    const cmdJson = {
+      redirect: {
+        url: addr2,
+      },
+    };
+    const redirect = api.Command.create(cmdJson);
+    sendFromServer(redirect, server1);
+
+    await server2.connected;
+
```
xyc, line 79: It seems that we don't enforce assertion in tests (https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/expect-expect.md#enforce-assertion-to-be-made-in-a-test-body-expect-expect), but it would be good practice to assert (expect) explicitly here.

src/client.ts
```diff
@@ -1263,6 +1286,10 @@ export class Client<Ctx = null> {
         chan0,
         error,
       });
+
+      if (this.redirectInitiatorURL) {
```
xyc, line 1290: Are we retrying connecting to both the initiator and the redirected server at the same time?

xyc: What if the initiator server has errored? `handleRedirect` does not have a backoff mechanism. It seems that we should utilize [retryConnect](https://github.com/replit/crosis/pull/137/files#diff-25d66d74617fe2e23d7946bd6e3ba95640ab1b9bc8947445d604fc271c7c1f12R1283-R1288) to connect to the initiator.

package.json
```diff
@@ -22,7 +22,9 @@
   "license": "MIT",
   "dependencies": {
     "abortcontroller-polyfill": "^1.7.3",
-    "engine.io-client": "^3.4.0"
+    "engine.io-client": "^3.4.0",
+    "jest-websocket-mock": "^2.3.0",
```
xyc, line 26: these mocks should be dev dependencies

src/__tests__/redirect.test.ts
brenoafb: :heavy_check_mark:

package.json
brenoafb: Good idea :heavy_check_mark:

src/client.ts
brenoafb: I'm thinking that we get retries 'for free', since `handleRedirect` calls open, which in turn calls connect.

xyc: just above this we still call `retryConnect`: https://github.com/replit/crosis/blob/cac7a11883cc075c3520c37a83766de52d3e4697/src/client.ts#L1283-L1288

is there a way to make the logic that connect to `redirectInitiatorURL` in `handleRedirect` to be part of `retryConnect`?

src/__tests__/redirect.test.ts
xyc: Jest run in band does this but afaik it's best for debugging https://archive.jestjs.io/docs/en/24.x/cli#--runinband

If you make a loop to check if the port is a duplicate it could be better, something like this https://github.com/replit/repl-it-web/pull/18016/files?diff=unified&w=0#diff-f97d48b055d87cb43ecbad5e2ee66eb6a684718f30694c5c0f9a0b494d50c516R183-R195. I'm not sure 
how `jest-websocket-mock` handles it but it seems safer to create ports that are all different from each other (using a Set to check for duplicate) or close the server after a test.

brenoafb: Took a quick look at how the library handles ports and it just checks whether there is an server listening by an address basis. We also can't use `:0` to let the OS pick a port.
At the end of the day I fixed it by just making sure servers are closed before finishing the test case.

src/client.ts
brenoafb: Ok, added a check in `retryConnect` to fallback to initiator after a maximum number of retries.

xyc (approved):

src/client.ts
xyc: sounds good, just to note that it could be minutes after 10 retries because of the exponential backoff

PR #141, opened 2022-08-12
classify retriable errors

xyc:
Why
===

Classify retriable and non-retriable errors so the client will not retry on unretriable errors, e.g. user errors.

What changed
============

Add error code for retriable and non-retriable error and logic to handle them

Test plan
=========

Automation and test the errors are not retriable for an unretriable error. e.g. the client doesn't send get connection metadata call in an expected situation of user error.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -6,6 +6,7 @@ import {
   FetchConnectionMetadataError,
   ConnectionState,
   FetchConnectionMetadataResult,
+  CloseCode,
 } from './types';
 import type {
   ConnectOptions,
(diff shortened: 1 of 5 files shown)


src/client.ts
```diff
@@ -1096,14 +1099,20 @@ export class Client<Ctx = null> {
...
-        const closeCodePolicyViolation = 1008;
-        if (closeEvent.code === closeCodePolicyViolation) {
+        if (closeEvent.code === CloseCode.POLICY_VIOLATION) {
           // This means that the token was rejected. We need to fetch another one.
           this.connectionMetadata = null;
         }
+
+        if (
+          closeEvent.code === CloseCode.USER_ERROR ||
+          closeEvent.code === CloseCode.FIREWALL_DENIED
```
lhchavez, line 1109: we should probably also come up with our own convention for future errors. like in addition to these two, all 44XX are non-retriable errors (similar to 4XX in HTTP) so that we don't need to coordinate beforehand.

lhchavez (approved):

src/client.ts
xyc: @lhchavez Makes sense. Where do we want to document this, in the README file of this repository? Noting here that 4001 TRY_ANOTHER_MACHINE is currently retriable so we would need to change it to another code.

PR #142, opened 2022-08-25
Make it possible to supply own backoff function

masad-frost:
Why
===
We should be able to easily tweak from the frontend

Stepping stone for #121

We should consider pulling in back this package into our monorepo until we're ready to open our API again.

What changed
============
Added option to supply a backoff function `getNextRetryDelayMs`, fallsback to current one.

Test plan
=========

Testing in the web repo, currently aiming to ship for incident response


masad-frost:
@replit/micromanager merge incident response, will test on main

masad-frost:
integration tests might fail due to incident, will try to isolate locally

PR #145, opened 2022-11-17
Add better error messages for non-retriable errors

lhchavez:
Why
===

Previously the concurrent Repl limit error was retriable, leading to a massive spam of reconnect attempts. This almost melted our infrastructure.

What changed
============

This change marks the concurrent Repl limit error as non-retriable and also adds nice error messages for the other non-retriable errors.

Test plan
=========

After this lands, the concurrent Repl  limit error is no longer retriable.


src/client.ts
```diff
@@ -1101,19 +1102,28 @@ export class Client<Ctx = null> {
...
         const closeEvent = <CloseEvent>event;
         if (closeEvent.code === CloseCode.POLICY_VIOLATION) {
-          // This means that the token was rejected. We need to fetch another one.
+          // This means that the token was rejected. Even though this is a
+          // permanent error from the perspective of the infrastructure, most
+          // of the time it can be corrected by fetching a new token, since
+          // this happens during cluster transfers most of the time (although
+          // sometimes it also happens when a particular Repl is tombstoned).
+          // TODO: Make the tombstone case return an USER_ERROR to have more
+          // clarity about why the Repl couldn't run.
```
cbrewster, line 1111:
```suggestion
          // sometimes it also happens when a particular Repl is taken down).
          // TODO: Make the takedown case return an USER_ERROR to have more
          // clarity about why the Repl couldn't run.
```

nit: just to differentiate between tombstone for cluster transfer and tombstone for takedown

cbrewster (approved):

lincoln-replit (approved):

src/client.ts
```diff
@@ -1101,19 +1102,28 @@ export class Client<Ctx = null> {
...
-
-        if (
-          closeEvent.code === CloseCode.USER_ERROR ||
-          closeEvent.code === CloseCode.FIREWALL_DENIED
-        ) {
+        } else if (closeEvent.code === CloseCode.USER_ERROR) {
+          errorMessage = 'Repl not allowed to run at this time. Please try again later.';
+          retriable = false;
+        } else if (closeEvent.code === CloseCode.FIREWALL_DENIED) {
+          errorMessage = "Can't connect to unfirewalled repl from firewall mode";
```
gburtini, line 1117: s/repl/Repl/g.

PR #146, opened 2022-11-17
v9.1.2

lhchavez:
Why
===

New release time!

What changed
============

* Bumped the version
* Bumped the version of the protocol so that the `firewallDenied` message was there.

Test plan
=========

:ship:.


yarn.lock
```diff
@@ -6851,7 +6851,7 @@ protobufjs@^6.10.2:
     "@protobufjs/pool" "^1.1.0"
     "@protobufjs/utf8" "^1.1.0"
     "@types/long" "^4.0.1"
-    "@types/node" "^13.7.0"
+    "@types/node" ">=13.7.0"
```
xyc, line 6854: Do we want to use a caret version here?

xyc (approved):

PR #149, opened 2022-11-27
enhancement(helper.ts): Doc-string implemented

pneb:
Why
===

Recommendation

What changed
============

The code now includes a doc-string.

Test Plan
=========

So that the user fully understands how to use the function.

### DIFF
--- a/src/util/helpers.ts
+++ b/src/util/helpers.ts
@@ -13,6 +13,9 @@ export function defaultGetNextRetryDelay(retryNumber: number): number {
 
   return Math.min(backoff, MAX_BACKOFF) + randomMs;
 }
+/**
+ * Check if supplied param is a standard WebSocket instance.
+ */
 
 function isWebSocket(w: unknown): w is WebSocket {
   if (typeof w !== 'object' && typeof w !== 'function') {


src/util/helpers.ts
```diff
@@ -13,6 +13,9 @@ export function defaultGetNextRetryDelay(retryNumber: number): number {
 
   return Math.min(backoff, MAX_BACKOFF) + randomMs;
 }
+/**
```
lhchavez, line 16:
```suggestion

/**
```

for consistency with other docstrings, there should be one empty line before

src/util/helpers.ts
```diff
@@ -13,6 +13,9 @@ export function defaultGetNextRetryDelay(retryNumber: number): number {
...
+ * Check if supplied param is a standard WebSocket instance.
+ */
```
lhchavez, line 19:
```suggestion
```

similarly, for consistency there shouldn't be any empty lines between the docstring and the function.

pneb: Alright, my bad. Thanks for fixing it. 🙂

PR #152, opened 2023-02-06
Add `ClientCloseReason.Temporary`.

gburtini:
Why
===

https://github.com/replit/repl-it-web/pull/27326

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -27,8 +27,7 @@ enum ClientCloseReason {
    */
   Intentional = 'Intentional',
   /**
-   * Called `client.close`, but we're going to try to reconnect
-   * at the client's convenience.
+   * Called `client.close`, but we're going to try to reconnect.
    */
   Temporary = 'Temporary',
   /**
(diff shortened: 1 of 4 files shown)


masad-frost (approved):
Can you make the call to `close` in `destroy` say it's not temporary?

In general, I don't feel good about `close` having that option AND also having separate functions for `close` and `destroy`. I think `close` should be always treated as temporary and `destroy` as permanent.

Another note is that if someone decides to temporarily close, but somewhere along the way they don't actually reconnect and decide to destroy the client, how do we inform the channel openers to clean up their resources.

src/client.ts
```diff
@@ -1688,7 +1714,7 @@ export class Client<Ctx = null> {
     ws.onclose = null;
     ws.onopen = null;
 
-    // Replace exististing error handler so an error doesn't get thrown.
+    // Replace existing error handler so an error doesn't get thrown.
```
masad-frost, line 1717: rofl

PR #155, opened 2023-02-24
(typos)

gburtini:
Why
===

Just want to clean out these typos before doing the 10.x branch.

What changed
============

Literally just typos. Since one was in an `Error` (which doesn't appear to be relied on anywhere), I bumped the patch.

Test plan
=========

:eyes: I inspected web for reliance on the _containter_ typo.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@replit/crosis",
-  "version": "9.2.0",
+  "version": "9.2.1",
   "description": "Goval connection and channel manager",
   "files": [
     "/dist"
--- a/src/client.ts
+++ b/src/client.ts
@@ -164,7 +164,7 @@ export class Client<Ctx = null> {
 
   /**
    * The connection might require multiple retries to be established.
-   * Anytime we need to retry, we should also add an incremental backoff,
+   * Anytime we need to retry, we should also add an incremental back off,
    * we do that using `setTimeout`. When the client closes before our
    * retry is initiated, we clear this timeout.
    *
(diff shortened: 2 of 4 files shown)


lhchavez (approved):

src/client.ts
```diff
@@ -164,7 +164,7 @@ export class Client<Ctx = null> {
 
   /**
    * The connection might require multiple retries to be established.
-   * Anytime we need to retry, we should also add an incremental backoff,
```
lhchavez, line 167: this one is correct, the word is "[backoff](https://en.wiktionary.org/wiki/backoff)".

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@replit/crosis",
-  "version": "9.2.0",
+  "version": "9.2.1",
```
masad-frost, line 3: Generally shouldn't bump versions in feature branches, it usually leads to conflicts and git tag confusion (usually do rebase merge for version bump branches, or just push to main directly)

src/client.ts
gburtini: cSpell disagrees. :-( I'll revert it though, I agree.

package.json
masad-frost: it's a bit pedantic since you're probably the only one with a feature branch haha

gburtini: Right, it is a race with other committers. I do not have permission to push to main though. :-)

src/client.ts
```diff
@@ -164,7 +164,7 @@ export class Client<Ctx = null> {
...
+   * Anytime we need to retry, we should also add an incremental back off,
```
gburtini, line 167:
```suggestion
   * Anytime we need to retry, we should also add an incremental backoff,
```

package.json
masad-frost: Seems like we're missing the git tag for 9.2.0 also

gburtini: ([shhhh](https://github.com/replit/crosis/releases/tag/9.2.0))

masad-frost: :shushing_face:

PR #156, opened 2023-02-25
[major->10.0.0] `openChannel` to always have a valid channel

gburtini:
Why
===

https://github.com/replit/repl-it-web/pull/28333

Includes #157 now.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -81,7 +81,14 @@ export class Client<Ctx = null> {
    * or the client closed permanently. Otherwise it'll be
    * CONNECTED or CONNECTING
    */
-  public connectionState: ConnectionState;
+  private connectionState: ConnectionState;
+
+  /**
+   * A list of listeners for connection state changes.
+   *
+   * @hidden
+   */
+  private connectionStateChangeFuncs: Array<(state: ConnectionState) => void>;
 
   /**
    * The websocket used for communication with the container.
@@ -228,6 +235,7 @@ export class Client<Ctx = null> {
     this.chan0Cb = null;
     this.chan0CleanupCb = null;
     this.connectionState = ConnectionState.DISCONNECTED;
+    this.connectionStateChangeFuncs = [];
     this.debugFuncs = [];
     this.bootStatusFuncs = [];
     this.userUnrecoverableErrorHandler = null;
@@ -411,7 +419,7 @@ export class Client<Ctx = null> {
 
     this.channelRequests.push(channelRequest);
 
-    if (this.connectionState === ConnectionState.CONNECTED && !sameNameChanRequests.length) {
+    if (this.getConnectionState() === ConnectionState.CONNECTED && !sameNameChanRequests.length) {
       // If we're not connected, then the request to open will go out once we're connected.
       // If there are channels with the same name then this request is queued after the other
       // channel(s) with the same name is done closing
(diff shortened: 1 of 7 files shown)


src/client.ts
```diff
@@ -1656,19 +1640,6 @@ export class Client<Ctx = null> {
...
-    } else if (!willClientReconnect) {
-      if (this.chan0Cb) {
-        this.chan0Cb({
-          channel: null,
-          error: new Error('Failed to open'),
-          context: this.connectOptions ? this.connectOptions.context : null,
-        });
-      } else if (closeResult.closeReason !== ClientCloseReason.Error) {
-        // if we got here as a result of an error we're not gonna call onUnrecoverableError again
-        this.onUnrecoverableError(new Error('open should have been called before `handleClose`'));
```
masad-frost, line 1668: Why get rid of this assertion? Does it not still make sense?

masad-frost (approved):
Looks good to me, just 1 q about an invariant assertion

src/client.ts
gburtini: Nope, this is wrong; I'll put it back.

I struggled with the decision a bit but with it drawn to my attention, it is clear it should stay.

gburtini: Can you look again? 

Before I merge this I'm going to open a PR into it to expose the statuses too.

src/client.ts
```diff
@@ -1656,19 +1640,9 @@ export class Client<Ctx = null> {
...
-
-        return;
-      }
+    } else if (!this.chan0Cb && closeResult.closeReason !== ClientCloseReason.Error) {
```
masad-frost, line 1643: For some reason !willClientReconnect was part of the invariance check, not sure what to make of it. Lemme spelunk in the history

masad-frost: In #61 we moved it from https://github.com/replit/crosis/pull/61/files#diff-25d66d74617fe2e23d7946bd6e3ba95640ab1b9bc8947445d604fc271c7c1f12L998-L1002 to https://github.com/replit/crosis/pull/61/files#diff-25d66d74617fe2e23d7946bd6e3ba95640ab1b9bc8947445d604fc271c7c1f12R1139. 

From what I can tell it was moved because that's the location this.chan0Cb was used in the handleClose and there was no other reason to the invariant assertion otherwise. Nothing to do with `!willClientReconnect` itself.

Feels like this check belongs to the top of the `handleClose` function tbh

masad-frost: Oh and one more thing, we should probably return, not fallthrough just to avoid cascading problems

gburtini: Hmm, I think I want to put back the willReconnect check. 

I have to check more carefully, but we might be able to get here when we're reconnecting from a involuntary disconnect in the wrong order of open/close.

masad-frost: Can we do a different error string when !willReconnect is different so we can understand when each might happen in sentry? We can move it to the beginning of the function in a future change.

It feels like I wasn't paying attention that it's a nested condition when I added it lol.

gburtini: Yep, ok.

gburtini: Bifurcated the error message. On more investigation, I no longer think I can get there in the way described above. We can ship like this.

gburtini:
Going to link this up one more time before doing the squash+publish; make sure it actually meets the needs of https://github.com/replit/repl-it-web/pull/28333 https://github.com/replit/repl-it-web/pull/28349

PR #157, opened 2023-02-25
[major->10.0.0] Expose connection state directly.

gburtini:
Why
===

https://github.com/replit/repl-it-web/pull/28349

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -81,7 +81,14 @@ export class Client<Ctx = null> {
    * or the client closed permanently. Otherwise it'll be
    * CONNECTED or CONNECTING
    */
-  public connectionState: ConnectionState;
+  private connectionState: ConnectionState;
+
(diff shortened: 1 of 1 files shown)


src/client.ts
```diff
@@ -805,6 +813,47 @@ export class Client<Ctx = null> {
...
+  /**
+   * Gets the current connection state.
+   *
+   * The listener functions are only called when the connection state changes.
+   * This function can be used to get the current state ahead of setting up a
+   * listener.
+   *
+   * @returns the current ConnectionState
+   * */
+
```
gburtini, line 854:
```suggestion
```

masad-frost (commented):
Lgtm

masad-frost (commented):
I feel like observable values are everywhere in JS there should be a common abstraction/standard for it

PR #159, opened 2023-02-26
Bump @replit/protocol to 0.2.93

gburtini:
Why
===

When we link to web with an out of date protocol, bad things happen. Crosis _supports_ the old protocol and the semver looks reasonable, so I did not update the package.json, just the lockfile.


lhchavez (approved):
is there a downside to also bumping `package.json`?

gburtini:
It is just semantically meaningful and I don't know how third parties are relying on this package. 

I'll just do it, probably any users are forced to update protocol in ways that aren't captured in semver anyway.

gburtini:
Oh, actually, the reason I didn't do it is that they're listed as `>=` rather than flexible (`^` or `~`). The current listed dependency is ostensibly correct and keeping the `>=` but adjusting the version will prevent users from using old, but still compatible, protocol definitions.

I  _think_ it doesn't matter for the reason in my last comment, but it won't be backwards compatible, so I'm going to punt for now. Changing the lockfile gets us the benefit we need for web.

PR #161, opened 2023-02-26
[minor->10.1.0] Defer close on all user-specified handleConnect events.

gburtini:
Why
===

[As part of 10.0.0](https://github.com/replit/crosis/pull/157), I exposed an observable on `connectionState` for https://github.com/replit/repl-it-web/pull/28349 creating a new place where a listener could call `client.close`. This means we have to defer that close event too.

So, I moved all that related logic inside `handleConnect`. It was only called in the one place anyway.

Also, clear listeners on destroy.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -1543,7 +1543,8 @@ export class Client<Ctx = null> {
     this.ws.onclose = onClose;
     this.ws.onerror = onClose;
 
-    // defer closing if the user decides to call client.close inside a callback.
+    // defer closing until the full connect sequence is complete
+    // in case the user decides to call client.close inside a callback.
     const originalClose = this.close;
     this.close = (args) =>
       setTimeout(() => {
@@ -1551,8 +1552,8 @@ export class Client<Ctx = null> {
       }, 0);
 
     // connection state possibly has a listener, so it needs the deferred close.
-    // note that CONNECTED fires _before_ the chan0Cb to match the
-    // original behavior of state being CONNECTED inside the chan0Cb.
+    // note that CONNECTED is set _before_ the chan0Cb to match the
+    // pre-10.1 behavior of state being CONNECTED inside the chan0Cb.
     this.setConnectionState(ConnectionState.CONNECTED);
     this.debug({ type: 'breadcrumb', message: 'connected!' });
 
@@ -1560,7 +1561,6 @@ export class Client<Ctx = null> {
       this.requestOpenChannel(channelRequest);
     });
 
-    // chan0Cb definitely has a callback.
     this.chan0CleanupCb = this.chan0Cb({
       channel: chan0,
       context: this.connectOptions.context,
--- a/src/client.ts
+++ b/src/client.ts
@@ -1543,10 +1543,6 @@ export class Client<Ctx = null> {
     this.ws.onclose = onClose;
     this.ws.onerror = onClose;
 
-    this.channelRequests.forEach((channelRequest) => {
-      this.requestOpenChannel(channelRequest);
-    });
-
     // defer closing if the user decides to call client.close inside a callback.
     const originalClose = this.close;
     this.close = (args) =>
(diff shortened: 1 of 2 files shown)


gburtini:
Also, this bumps the minor and will be published, it will be the first build to include all those version bumps from @dependabot.

src/client.ts
```diff
@@ -758,6 +758,7 @@ export class Client<Ctx = null> {
     }
 
     this.debug = () => {};
+    this.connectionStateChangeFuncs = [];
```
masad-frost, line 761: Oops

gburtini: Yeah, bleh.

src/client.ts
```diff
@@ -1560,9 +1543,30 @@ export class Client<Ctx = null> {
     this.ws.onclose = onClose;
     this.ws.onerror = onClose;
 
+    // defer closing if the user decides to call client.close inside a callback.
+    const originalClose = this.close;
+    this.close = (args) =>
```
masad-frost, line 1548: I hate this lol

masad-frost (approved):

src/client.ts
gburtini: Yeah, we knew it was going to cause problems and here we are...

gburtini: I suppose we could make it happen all in the same execution path by setting a sentinel instead?

masad-frost (commented):
There's a test for close in chan0cb I think if you wanna duplicate that for state

src/client.ts
```diff
@@ -1560,9 +1543,30 @@ export class Client<Ctx = null> {
...
+    // connection state possibly has a listener, so it needs the deferred close.
+    // note that CONNECTED fires _before_ the chan0Cb to match the
+    // original behavior of state being CONNECTED inside the chan0Cb.
+    this.setConnectionState(ConnectionState.CONNECTED);
+    this.debug({ type: 'breadcrumb', message: 'connected!' });
+
     this.channelRequests.forEach((channelRequest) => {
       this.requestOpenChannel(channelRequest);
     });
+
```
masad-frost, line 1562: Btw we used to do the chan0cb before other channels, should we keep it that way?

gburtini: Yes, probably. Let me look again.

gburtini: Oh yeah, I got it backwards for sure. Will fix.

src/client.ts
```diff
@@ -1560,9 +1543,30 @@ export class Client<Ctx = null> {
     this.ws.onclose = onClose;
     this.ws.onerror = onClose;
 
+    // defer closing if the user decides to call client.close inside a callback.
```
masad-frost, line 1546: This comment is so bad, "why not what" lol

masad-frost: #55 Says we used to end up in a "nasty place"

src/client.ts
gburtini: Fixed this one.

src/client.ts
gburtini: We can revise the comment, but I'm currently planning to leave the implementation how it was (ugly deferred close) because changing it gets caught up on something in the `ws.onclose` loop.

src/client.ts
```diff
@@ -1560,9 +1543,31 @@ export class Client<Ctx = null> {
...
+        originalClose(args);
+      }, 0);
+
+    // connection state possibly has a listener, so it needs the deferred close.
+    // note that CONNECTED is set _before_ the chan0Cb to match the
+    // pre-10.1 behavior of state being CONNECTED inside the chan0Cb.
+    this.setConnectionState(ConnectionState.CONNECTED);
+    this.debug({ type: 'breadcrumb', message: 'connected!' });
+
+    // chan0Cb definitely has a callback.
```
gburtini, line 1560:
```suggestion
```

gburtini:
Huh, reordering to put chan0 first seems to have broken a test. :-(

src/client.ts
```diff
@@ -1295,39 +1296,7 @@ export class Client<Ctx = null> {
...
-            return;
-          }
-
-          if (!this.chan0Cb) {
-            this.onUnrecoverableError(new Error('Expected chan0Cb to be truthy'));
-
-            return;
-          }
-
-          this.handleConnect();
```
gburtini, line 1316: Wait, @masad-frost I think handleConnect used to be called _before_ the chan0Cb, no?

src/client.ts
gburtini: Ha, no, I think I had this right before!

gburtini: The path here (before and now) is that chan0 opens, which sets off this connection chain, we requestOpenChannel for each channel request _and then_ call the chan0Cb.

Changing that order breaks a test related to reopening channels with the same name and breaks workspace. I'm not quite sure why, but I want to preserve the original order.

PR #163, opened 2023-03-01
[10.1.1] Add breadcrumbs for out of sync `channelRequests` and `channels` at cleanup time.

gburtini:
Why
===

https://github.com/replit/repl-it-web/pull/28555
https://replit-kq.sentry.io/issues/3965488037

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -1680,24 +1680,26 @@ export class Client<Ctx = null> {
     }
 
     if (Object.keys(this.channels).length !== 0) {
+      // this should never happen, because the channelRequests should have
+      // triggered cleanup of all this, so dump a bunch of breadcrumbs to
+      // help chase down how we got here.
+      for (const key in this.channels) {
+        const channel = this.channels[key];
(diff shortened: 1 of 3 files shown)


src/client.ts
```diff
@@ -1678,6 +1683,22 @@ export class Client<Ctx = null> {
       this.channels = {};
       if (closeResult.closeReason !== ClientCloseReason.Error) {
         // if we got here as a result of an error we're not gonna call onUnrecoverableError again
+        // but we are going to dump a bunch of breadcrumbs to help chase down how we got here.
+
+        for (const key in this.channels) {
+          const channel = this.channels[key];
```
masad-frost: nit 

```suggestion
        for (const channel in Object.values(this.channels)) {
```

also avoids https://eslint.org/docs/latest/rules/guard-for-in

masad-frost (approved):

src/client.ts
gburtini: OK, sort of done.

PR #164, opened 2023-03-04
[10.1.2] Even more logs.

gburtini:
Why
===
Follow up to https://github.com/replit/crosis/pull/163

https://github.com/replit/repl-it-web/pull/28739

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -1737,15 +1737,12 @@ export class Client<Ctx = null> {
       }
 
       this.channels = {};
-      if (closeResult.closeReason !== ClientCloseReason.Error) {
-        // if we got here as a result of an error we're not gonna call onUnrecoverableError again
-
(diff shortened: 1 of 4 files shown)


src/client.ts
```diff
@@ -1737,15 +1737,12 @@ export class Client<Ctx = null> {
...
 
       this.channels = {};
-      if (closeResult.closeReason !== ClientCloseReason.Error) {
-        // if we got here as a result of an error we're not gonna call onUnrecoverableError again
-
-        this.onUnrecoverableError(
-          new Error('channels object should be empty after channelRequests and chan0 cleanup'),
-        );
-
-        return;
```
gburtini, line 1706: Note that I let us fall through here; I think it is safe.

masad-frost (approved):

PR #165, opened 2023-03-06
[10.1.2] Help Sentry bucket channel errors more usefully.

gburtini:
Why
===

We often lose the stack trace here. This adds the service name when it is present so we can at least see which service is misbehaving.


mikewesthad (approved):
Ah smart!

gburtini:
(@masad-frost this is unrelated to the intent of the #164 changes, but [I think it will help us understand the scope of these](https://replit.slack.com/archives/C030RJ2PRC7/p1678113377200969) and I don't see a good reason not to ship at the same time)

masad-frost (approved):

PR #167, opened 2023-03-07
[10.2.0] Defer all connection listeners to outside of our synchronous context.

gburtini:
Why
===

A bunch of unexpected behavior was introduced by mapping container and Crosis state in https://github.com/replit/crosis/pull/161 because it allowed consumers of this API to make changes _during_ synchronous events.

This prevents that by dumping to the next tick.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -874,7 +874,15 @@ export class Client<Ctx = null> {
    */
   private setConnectionState = (connectionState: ConnectionState): void => {
     this.connectionState = connectionState;
-    this.connectionStateChangeFuncs.forEach((f) => f(connectionState));
+    setImmediate(() => {
+      // We need to defer these so that the connection state is announced to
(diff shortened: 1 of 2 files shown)


src/client.ts
```diff
@@ -874,7 +874,15 @@ export class Client<Ctx = null> {
    */
   private setConnectionState = (connectionState: ConnectionState): void => {
     this.connectionState = connectionState;
-    this.connectionStateChangeFuncs.forEach((f) => f(connectionState));
+    setImmediate(() => {
```
masad-frost, line 877: we don't have `setImmediate` in the browser context. The closest thing is probably `Promise.resolve().then(callback)`, `setTimeout(callback, 0)` is another option

gburtini: I put these back to 0ms `setTimeout`.

masad-frost (approved):

PR #168, opened 2023-03-08
[10.3.0] Restore out of sync channels invariant.

gburtini:
Why
===

The underlying bug I was chasing down is fixed in #167, we made this non-fatal to suppress the pain workspace-side, but I want to undo all the extra handlers now that this should really mark a significant error. [See resolved error](https://replit-kq.sentry.io/issues/3981034175/?_allp=1&query=channels+out+of+sync&referrer=issue-stream&statsPeriod=14d)

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -1808,10 +1808,12 @@ export class Client<Ctx = null> {
       }
 
       this.channels = {};
-      this.debug({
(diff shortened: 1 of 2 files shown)


moudy (approved):
🧹

PR #170, opened 2023-03-08
[10.3.0] If the server gives us two channels with the same ID, explode.

gburtini:
Why
===

I added this check while diagnosing the out of sync channels issue. It is the opposite of that issue though, we just may end up with multiple channel requests that map to the same channel.

I don't trust my understanding of the protocol well enough to know that this should always be fatal; it doesn't occur in my testing nor the test suite, but I'll tag Faris for this review.

If there's a protocol reason not to ship this, I'll propose removing the invariant entirely as its purpose was just diagnostic.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -588,20 +588,9 @@ export class Client<Ctx = null> {
       }
 
       if (this.channels[id] != null) {
-        const serviceName =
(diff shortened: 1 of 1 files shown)


masad-frost (approved):
Should never happen, yes

PR #171, opened 2023-03-08
[10.3.0] Cleanup and standardize types and names of remaining breadcrumbs.

gburtini:
Why
===

I took an opinionated first pass at cleaning these up as part of undoing the unnecessary parts of my 10.0 - 10.2 diagnostic work. This required a little bit of unexpected web-side changes :thinking:, which is done here:  https://github.com/replit/repl-it-web/pull/28977

- Conceptually grouped all breadcrumbs, this should make them easier to filter on in the Sentry UI.
- Renamed ones that were close-but-not-correct mappings to existing function names.
- Always used the derived service/name types; Sentry correctly handles nulling out the thunks anyway.
- Reordered the types file a little so it was more obvious where to put things.

Very open to feedback here, I'm sure there's a better way to do this, I'm just unaware of some of the context of these logs (e.g., maybe the fallback/redirect logs should actually be removed at this point as they don't look to offer a lot of diagnostic value).

### DIFF
--- a/src/__tests__/index.test.ts
+++ b/src/__tests__/index.test.ts
@@ -166,7 +166,7 @@ concurrent('client retries and caches tokens', (done) => {
 
   let reconnectCount = 0;
   client.onDebugLog((log) => {
-    if (log.type === 'breadcrumb' && log.message === 'client closed') {
+    if (log.type === 'breadcrumb' && log.message === 'status:closed') {
       expect(fetchConnectionMetadata).toHaveBeenCalledTimes(1);
       expect(onConnect).not.toHaveBeenCalled();
       done();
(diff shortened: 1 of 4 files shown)


gburtini:
lol @masad-frost i tried to save you from this one with the rotation...

masad-frost (approved):
Looks fine to me :shrug: 

The only thing is that we have some metrics on datadog that use these messages verbatim, so you're gonna have to hunt those down and add the new messages as an alternative source. Or remap them back to the old messages before sending them to datadog.

gburtini:
How do they even get to Datadog? Is that just the code I updated here: https://github.com/replit/repl-it-web/pull/28977

gburtini:
I merged this one to the wrong place too _and then published 11_, what am I doing.

PR #174, opened 2023-03-14
[10.3.0] Smarter closed channel error text.

gburtini:
Why
===

Getting undefined service channels, I assume it is chan0, because no other call does that, but let's clean it up for Sentry anyway.

### DIFF
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -15,7 +15,7 @@ export class Channel {
   /**
    * The name of the service associated with the channel.
    */
-  public readonly service?: string;
+  public readonly service: string;
 
   /**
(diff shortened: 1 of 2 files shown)


masad-frost (approved):
Can we instead make the `service` a required field and pass that as `channel0` when we create it?

gburtini:
Done, but it intimidates me a bit. The Channel class does nothing useful with its service name (instead, a request by the Client on chan0 does that), so I suppose it is fine?

masad-frost:
Yeah should be fine. I remember adding those purely for debugging purposes. But...future us might decide to use it lol

PR #172, opened 2023-03-08
[10.3.0] Version bump

gburtini:
Why
===

Changes included in this version:

- #168 
- #169 
- #170 
- #171 

I'm open to disagreement that changing breadcrumb names is a minor bump; in which case, we go to 11 instead.

PR #175, opened 2023-03-16
[11.0.0] Actual version bump.

gburtini:
I merged the version bump into a dead/merged old branch by mistake. PR had a bad name anyway, since we decided it was a major bump.

Details here: https://github.com/replit/crosis/pull/172

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@replit/crosis",
-  "version": "10.2.0",
+  "version": "11.0.0",
   "description": "Goval connection and channel manager",
   "files": [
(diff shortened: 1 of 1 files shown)


lunaroyster (approved):
ooh a major!

btw remember to also update the version in the lockfile by running the install command

gburtini:
This is Crosis. The library lockfile should be unaware of its own file (but web-side, yep).

lunaroyster:
ohhh this is a difference between yarn and npm

npm package-lock.json files have their own package's version in them (which is *really* annoying when they go out of date)

PR #176, opened 2023-03-17
[bsod] Abort the fetch metadata request if it is ongoing.

gburtini:
Why
===

I believe this is the root cause of [this error](https://replit-kq.sentry.io/issues/3987472868/?project=6229092&query=is%3Aunresolved&referrer=issue-stream&sort=user&statsPeriod=14d) and possibly all our other cases where the AbortController should have been fired but was not.

Linked error results when `chan0` is closed when we call the `onCommand` [during open](https://github.com/replit/crosis/blob/286996a3c210565aaacc034a9a24b5269994bf48/src/client.ts#L1318). The only way for this to happen is for the channel to close during the fetch metadata event.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -1477,6 +1477,14 @@ export class Client<Ctx = null> {
           wsReadyState: this.ws ? this.ws.readyState : undefined,
         },
       });
+
+      if (this.fetchTokenAbortController) {
+        // this can happen while we're fetching the token. if so, when we
+        // close chan0, we need to know we're aborted for that request.
+        this.fetchTokenAbortController.abort();
+        this.fetchTokenAbortController = null;
+      }
+
       chan0.handleClose({ initiator: 'client', willReconnect: true });
       delete this.channels[0];
 


masad-frost (changes requested):
It's really fucked that we ever get into a state like this. We get here after `retryConnect` is called [here](https://github.com/replit/crosis/blob/23a18f398ec5185bde538ec9f76558512499ae70/src/client.ts#L1398) or [here](https://github.com/replit/crosis/pull/176/files) both of which happen after `fetchTokenAbortController` [is set to null](https://github.com/replit/crosis/blob/23a18f398ec5185bde538ec9f76558512499ae70/src/client.ts#L1082)

Maybe we have interlacing connects or something? Can we find an invariant check we can put higher up in this stack?

masad-frost:
Also, sorryt for the hold up

gburtini:
Hmm, agreed there looks to be a root-er cause. :-1:

gburtini:
I suspect this wasn't going to fix anyway. I don't have a fix at the moment, but here's what I found:

- We set up a `chan0` and call it open.
- We call `fetchConnectionMetadata` asynchronously.
- `chan0` is now closed.

The `retryTimeout` _explicitly_ closes `chan0`, but we have a log right before it, so we know it isn't called in 90% of the cases of the error. The other path is `handleClose` closing the channel. This _definitely is happening_.

I also inspected the `fetchConnectionMetadata` implementation on the other side to see if it could plausibly not return the Aborted state. There isn't.

TLDR: I suspect your root cause is correct and we somehow have interlaced connections. This is not the fix for that. Possibly related to reconnection monitor or ping-pong.

PR #177, opened 2023-03-17
[logs] #171 take two (Cleanup and standardize types and names of remaining breadcrumbs.)

gburtini:
Why
===

I merged #171 into an intermediary branch, which was already merged. I'm not clear why Github didn't automatically change my branch to main when that merged, but in any case, it did not get published.

We're ready for the change over here: https://github.com/replit/repl-it-web/pull/28977, but unfortunately will have to go to 12 now to ship.

TLDR: this has already been approved, I just merged it into a dead intermediary branch and need approval to merge to main.

PR #178, opened 2023-03-17
[12.0.0] Version bump.

gburtini:
Why
===

When I published 11, I screwed up a fraction of the merge path. Unfortunately, it means the breaking change (#171) was not included in what went out as 11.

So, here we are again. This time with a breaking change (171 take two) and a bug fix.

- #177 
- #176

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@replit/crosis",
-  "version": "11.0.0",
+  "version": "12.0.0",
   "description": "Goval connection and channel manager",
   "files": [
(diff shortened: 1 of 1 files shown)


bradymadden97 (approved):

gburtini:
In case someone comes along and needs to ship this before I get back to it, this needs to go out with it: https://github.com/replit/repl-it-web/pull/28977

gburtini:
Going to ship #179 and #180 with this instead because my approach for #176 didn't pan out.

PR #179, opened 2023-03-27
`Error` :handshake: `CrosisError`

gburtini:
Why
===

We're currently coercing structured data to strings to pass it from Crosis to web to Sentry. This allows us to add some structured data as `tags` or `extras` (aligned with the Sentry concepts) and is extensible for other metadata as necessary.

### DIFF
--- a/src/util/CrosisError.ts
+++ b/src/util/CrosisError.ts
@@ -15,4 +15,8 @@ export default class CrosisError extends Error {
     this.extras = extras;
     this.tags = tags;
   }
+
+  static fromError(error: Error): CrosisError {
+    return new CrosisError(error.message);
+  }
 }
--- a/src/util/CrosisError.ts
+++ b/src/util/CrosisError.ts
@@ -17,6 +17,9 @@ export default class CrosisError extends Error {
   }
 
   static fromError(error: Error): CrosisError {
-    return new CrosisError(error.message);
+    const crosisError = new CrosisError(error.message);
+    crosisError.stack = error.stack;
+
+    return crosisError;
   }
 }
--- a/src/util/CrosisError.ts
+++ b/src/util/CrosisError.ts
@@ -1,0 +1,18 @@
+export default class CrosisError extends Error {
+  /** extras - a map of event data in a format appropriate to send to an error logging service */
+  public readonly extras: Record<string, unknown> | null;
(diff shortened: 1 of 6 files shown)


gburtini:
Reviewers, note that `client.ts` is heavily changed and Github collapses it by default. You'll want to see it.

lunaroyster (approved):

src/util/CrosisError.ts
```diff
@@ -0,0 +1,22 @@
...
+  constructor(
+    message: string,
+
+    extras: Record<string, unknown> | null = null,
+    tags: Record<string, string> | null = null,
+  ) {
+    super(message);
+
+    this.extras = extras;
+    this.tags = tags;
```
masad-frost, line 16:
```suggestion
    super(message);

    this.name = 'CrosisError';
    this.extras = extras;
    this.tags = tags;
    
    if ('captureStackTrace' in Error) {
      Error.captureStackTrace(this, CrosisError);
    } else {
      this.stack = Object.defineProperty(this, 'stack', {
        enumerable: false,
        value: Error(message).stack,
        writable: true,
        configurable: true,
      });
    }
```

Most of the time the stack stuff is unnecessary, but iirc i had an issue with stack traces in the past when i didn't do that.

src/util/CrosisError.ts
```diff
@@ -0,0 +1,22 @@
...
+  }
+
+  static fromError(error: Error): CrosisError {
+    return new CrosisError(error.message);
+  }
```
masad-frost, line 21:
```suggestion
  static fromError(error: Error): CrosisError {
    const crosisError = new CrosisError(error.message);
    crosisError.stack = error.stack;
    return crosisError
  }
```

we need to preserve stack traces :scream:

bradymadden97 (approved):

src/util/CrosisError.ts
gburtini: I guess, I think all this does is drop the `fromError` part of the stack trace (everything prior will be in the new stack), but that is probably helpful for Sentry.

src/util/CrosisError.ts
masad-frost: More importantly is that we set `name`

masad-frost: Yeah, maybe the stack stuff is not needed in this context

PR #180, opened 2023-03-27
Add some metadata to certain `CrosisError`s.

gburtini:
Why
===

- Trying to listen on closed channel: not a lot of additional debugging value here.
- Send while closed: we get the command, which hopefully lets us trace back the exact call in the service.
- Send while closing: we get the command, which hopefully lets us trace back the exact call in the service.
- `open` should've been called before `handleClose`: minor additive debug value, the connection state _might_ add some insight here, this feels related to #176.

### DIFF
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -114,6 +114,26 @@ export class Channel {
     };
   };
 
+  private wrapError = (e: CrosisError, cmd: api.ICommand): CrosisError => {
+    return new CrosisError(
+      e.message,
+      {
+        ...e.extras,
+        command: cmd,
+        // ref identifies if this was a request or a send.
+        commandRef: cmd.ref,
+        // keys retained separately to address potential privacy filtering.
+        commandKeys: Object.keys(cmd),
+        channelName: this.name,
+        channelId: this.id,
+      },
+      {
+        ...e.tags,
+        service: this.service,
+      },
+    );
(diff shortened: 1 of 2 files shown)


Monkatraz (approved):

src/channel.ts
```diff
@@ -198,4 +201,37 @@ export class Channel {
...
+
+  /**
+   * Append some extra information to an error, appropriate for being forwarded
+   * off to an error capture service.
+   *
+   * @param e - the error to wrap
+   * @param cmd  - the command that was being sent or received when the error occurred
+   * @returns the wrapped error
+   */
+  private wrapError = (e: CrosisError, cmd: api.ICommand | null = null): CrosisError => {
```
masad-frost, line 213:
```suggestion
  private createError = (message: string, cmd: api.ICommand | null = null): CrosisError => {
```

curious why not this?

Also I remember sentry giving me shit when i did this for OTClient, it grouped everything as if they're 1 error lol, hence [this fingerprinting logic](https://github.com/replit/repl-it-web/blob/2b17966f02dfb6b4644262f8df9ebc3d6881bb80/app/workspace/bsod/sentry.ts#L61). Maybe`this.getExtras` is more suitable.

gburtini: Suggestion LGTM.

PR #182, opened 2023-05-04
add priority to channel requests

moudy:
Why
===

Since requests are queued up until we have a connection less time sensitive channels might get priority over others. Adding a way to specify the priority lets the user prioritize certain channel requests  over others

What changed
============

Added an optional priority option to `channelRequest`

Test plan
=========

Added a basic test for the sorting. I think the real test will need to based on metrics from real usage

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -8,6 +8,7 @@ import {
   FetchConnectionMetadataResult,
   CloseCode,
   ClientCloseReason,
+  ChannelRequestPriority,
 } from './types';
 import type {
   ConnectOptions,
@@ -39,11 +40,6 @@ type CloseResult =
       error: CrosisError;
     };
 
-export enum ChannelRequestPriority {
-  High = 'High',
-  Low = 'Low',
-}
-
 type ChannelRequest<Ctx> =
   | {
       options: ChannelOptions<Ctx>;
@@ -52,7 +48,7 @@ type ChannelRequest<Ctx> =
       closeRequested: boolean;
       channelId: number;
       cleanupCb: ReturnType<OpenChannelCb<Ctx>>;
-      priority?: ChannelRequestPriority;
+      priority: ChannelRequestPriority;
     }
   | {
       options: ChannelOptions<Ctx>;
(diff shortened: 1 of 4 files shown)


src/client.ts
```diff
@@ -39,6 +39,11 @@ type CloseResult =
       error: CrosisError;
     };
 
+export enum ChannelRequestPriority {
+  High = 'High',
```
masad-frost, line 43: We could use number enums here to simplify the sorting logic

src/client.ts
```diff
@@ -47,6 +52,7 @@ type ChannelRequest<Ctx> =
       closeRequested: boolean;
       channelId: number;
       cleanupCb: ReturnType<OpenChannelCb<Ctx>>;
+      priority?: ChannelRequestPriority;
```
masad-frost, line 55: Can we set this to an explicit default value instead of undefined as an implicit lowest level value. Maybe add a medium level and set that as the default. I actually don't think the lowest should be the default because setting Low shows some intent to deprioritized, and if Low was higher than the default it'd be weird.

It'd make the sorting logic and readability of code much easier to grok and obvious.

src/client.ts
```diff
@@ -1620,7 +1659,7 @@ export class Client<Ctx = null> {
     this.setConnectionState(ConnectionState.CONNECTED);
     this.debug({ type: 'breadcrumb', message: 'status:connected' });
 
-    this.channelRequests.forEach((channelRequest) => {
+    [...this.channelRequests].sort(sortByPriority).forEach((channelRequest) => {
```
masad-frost, line 1662: I wonder if we should do this in a preprocessing step after we push to the array. Not sure about the trade-offs, but seems better to have it always be sorted from the get go.

masad-frost (approved):
Looks good. Left some loosely held opinions that are worth addressing imo

src/client.ts
moudy: good idea.  I like the strings for easier debugging but in the case being able to use them to sort is very convenient

src/client.ts
moudy: sg, I added a medium option

src/client.ts
moudy: updated to sort right after pushing

masad-frost (approved):
Lovely thank you

PR #185, opened 2023-06-14
Transactions.

gburtini:
Why
===

For reasons related to ergonomics of the API and realism about where and how it is being used. (I think this greatly cleans up a lot of our `channelClosed`/check closed stuff, misuse of which is a source of many visible/Sentry'd quality issues.)

What changed
============

Added `channel.transaction`, a method that lets you run a group of events with defined behavior around the failure mode. 

And `Channel::transaction` which, if both are to be retained, probably needs a more distinct name that can run more complex logic.

It is possible we'll only want to keep one of these or even escalate the concept to a single message or the container service.

Test plan
=========

Literally nothing and there's obvious(ish) bugs in the current take. This will change before the PR becomes non-draft. Just RFC for now.

### DIFF
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -182,6 +182,46 @@ export class Channel {
   };
 
   /**
+   * transaction([
+   *  { exec: { args: ['git', 'init'] }}
+   *  { exec: { args: ['git', 'add', '.'] }}
+   * ])
+   */
+  public transaction = async (
+    cmds: api.ICommand[],
+    behavior: TransactionBehavior,
+    // TODO: maybe async pattern options, like serialize/fire-and-forget
+  ): Promise<RequestResult[]> => {
+    const responses = [];
+    for (let i = 0; i < cmds.length; i++) {
+      const response = await this.request(cmds[i]);
+
+      if (response.channelClosed) {
+        if (behavior === 'retry') {
+          return this.transaction(cmds, behavior);
+        } else if (behavior === 'continue') {
+          return this.transaction(cmds.slice(i), behavior);
+        } else if (behavior === 'throw') {
+          throw new Error('Channel closed');
+        } else if (typeof behavior === 'function') {
+          return behavior(cmds, i, responses, response);
+        } else if (behavior === 'ignore') {
+          // the channel is closed, so we can't actually make any more requests
+          // so we just return the responses we have.
+
+          return responses;
+        }
+
+        throw new Error('Invalid behavior');
+      }
+
+      responses.push(response);
+    }
+
+    return responses;
+  };
+
+  /**
    * @hidden should only be called by [[Client]]
    *
    * Called when the channel receives a message
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -3,7 +3,7 @@ import type { ChannelCloseReason, RequestResult } from './types';
 import CrosisError from './util/CrosisError';
 
 type CustomFallbackBehavior = (
-  cmds: api.ICommand[],
+  commands: api.ICommand[],
   failedIndex: number,
   successfulResults: RequestResult[],
   failureResult: RequestResult,
@@ -258,22 +258,22 @@ export class Channel {
    * ])
    */
   public transaction = async (
-    cmds: api.ICommand[],
+    commands: api.ICommand[],
     behavior: TransactionBehavior | CustomFallbackBehavior,
   ): Promise<RequestResult[]> => {
     const responses = [];
-    for (let i = 0; i < cmds.length; i++) {
-      const response = await this.request(cmds[i]);
+    for (let i = 0; i < commands.length; i++) {
+      const response = await this.request(commands[i]);
 
       if (response.channelClosed) {
         if (behavior === 'retry') {
-          return this.transaction(cmds, behavior);
+          return this.transaction(commands, behavior);
         } else if (behavior === 'continue') {
-          return this.transaction(cmds.slice(i), behavior);
+          return this.transaction(commands.slice(i), behavior);
         } else if (behavior === 'throw') {
           throw new Error('Channel closed');
         } else if (typeof behavior === 'function') {
-          return behavior(cmds, i, responses, response);
+          return behavior(commands, i, responses, response);
         } else if (behavior === 'ignore') {
           // the channel is closed, so we can't actually make any more requests
           // so we just return the responses we have.
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -209,10 +209,15 @@ export class Channel {
     behavior: Omit<TransactionBehavior, 'continue'>,
     invariant = (_error: Error) => true,
 
+    retryArguments = {
+      maxRetries: -1,
+      retryFrequency: 1000,
+    },
+    retryCount = 0,
+  ): Promise<RequestResult | void> => {
     try {
-      // TODO: check status closed somehow?
-
       const result = await fn();
+
       return result;
     } catch (e) {
       if (!(e instanceof Error)) {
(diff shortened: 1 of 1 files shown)


src/channel.ts
```diff
@@ -172,6 +193,90 @@ export class Channel {
...
+   * transactions(() => {
+   *   const useResult = channel.whatever();
+   *
+   *   useResult.something();
+   *
+   *   channel.another();
+   * }, 'retry');
+   */
+  static transaction = async (
+    fn: () => Promise<RequestResult>,
```
jackyzha0, line 208:
```suggestion
    tx: () => Promise<RequestResult>,
```

and elsewhere. `fn` is pretty undescriptive

src/channel.ts
```diff
@@ -2,6 +2,27 @@ import { api } from '@replit/protocol';
 import type { ChannelCloseReason, RequestResult } from './types';
 import CrosisError from './util/CrosisError';
 
+type CustomFallbackBehavior = (
+  cmds: api.ICommand[],
+  failedIndex: number,
+  successfulResults: RequestResult[],
+  failureResult: RequestResult,
```
jackyzha0, line 9:
```suggestion
  failedResult: RequestResult,
```

keep grammar consistent :)

src/channel.ts
```diff
@@ -2,6 +2,27 @@ import { api } from '@replit/protocol';
...
+) => Promise<RequestResult[]>;
+
+type TransactionBehavior = 'retry' | 'continue' | 'throw' | 'ignore' | CustomFallbackBehavior;
```
jackyzha0, line 12:
```suggestion
/**
 * `'retry'`: Retries entire transaction if a single command fails
 * `'continue'`: Retries transaction starting from the failed command if any commands fail
 * `'throw'`: Bail on any failed command
 * `CustomFallbackBehavior`: custom function to handle failures
 */
type TransactionBehavior = 'retry' | 'continue' | 'throw' | 'ignore' | CustomFallbackBehavior;
```

src/channel.ts
gburtini: I don't think `tx` is more descriptive. I'll add a JSDoc?

jackyzha0: yeah sure

src/channel.ts
gburtini: Wouldn't I need to change it to succeededResults? I'll swap successful for success and then I think the consistency unambiguous.

src/channel.ts
```diff
@@ -172,6 +198,103 @@ export class Channel {
...
+   */
+  public transaction = async (
+    commands: api.ICommand[],
+    behavior: TransactionBehavior | CustomFallbackBehavior,
+  ): Promise<RequestResult[]> => {
+    const responses = [];
+    for (let i = 0; i < commands.length; i++) {
+      const response = await this.request(commands[i]);
+
+      if (response.channelClosed) {
```
masad-frost, line 273: don't we create a new channel when it closes?

src/channel.ts
```diff
@@ -2,6 +2,32 @@ import { api } from '@replit/protocol';
...
+  failedIndex: number,
+  successResults: RequestResult[],
+  failureResult: RequestResult,
+) => Promise<RequestResult[]>;
+
+/**
+ * `'retry'`: Retries entire transaction if a single command fails
+ * `'continue'`: Retries transaction starting from the failed command if any commands fail
+ * `'throw'`: Bail on any failed command
+ */
```
masad-frost, line 16: Have we confirmed some places where we can use these as is? For example, I imagine most places you have to confirm some remote state after reconnecting before you decide what you need to do.

The places where we blindly continue/retry, like the FS service, are technically wrong but exist for ergonomics of the original caller, usually the UI, and if we're being honest, because I didn't have time/was too lazy to go deep on proper retry logic, which will never be correct unless the FS service is very stateful and the protocol supports advanced collaboration techniques.

I'd at least rather not codify it and make it easy to do something that's incorrect.

src/channel.ts
masad-frost: Yeah this happens after we reconnect. 

https://github.com/replit/crosis/blob/main/src/client.ts#L602-L608

I think the thinking Moudy and I had was to force people to get a new handle and run through the OG callback again. When we were reusing channels we found ourselves in weird undefined states often.

gburtini: Yup. Probably have to remove this idea (or migrate it to the container service).

gburtini: This particular take isn't as ergonomic as the other one anyway, for example, it doesn't let you do anything with the intermediary responses and doesn't fit nicely into how we currently generate/expose ICommands.

src/channel.ts
gburtini: I'll address the substance of your comment, but do you have an immediately actionable suggestion or is this an assertion that you'd prefer to leave the usage how it is?

masad-frost: Yeah I'm not saying we don't do anything, my position is that this may be premature and potentially problematic. Let me add some more thoughts before suggesting anything.

I feel like every situation that requires a feature like channel.transaction needs to handle channel closures in a specific way. This introduces an element of variability that our proposed change might not cover entirely. Depending on the context, we may need to handle channel closures differently and depending on where we are in the "transaction".

I also feel like this is a higher level concept than the protocol layer here in crosis. It maybe more context or service specific.

The changes, as they stand now, could make it easier to write incorrect implementations, especially around retry logic. By providing a generic mechanism to handle a group of events with defined behavior around the failure mode, we might be leading developers into a false sense of security. The resulting code might work in some cases, but fail in edge cases where the context-specific handling of channel closures was required.

src/channel.ts
```diff
@@ -172,6 +198,103 @@ export class Channel {
...
+
+      return result;
+    } catch (e) {
+      if (!(e instanceof Error)) {
+        throw e;
+      }
+
+      if (!invariant(e)) {
+        throw new InvariantViolation(e);
+      }
```
gburtini, line 234: Possible we could address Faris's point by checking the invariant prior to retry, rather than as an assertion on the form of the error.

src/channel.ts
masad-frost: Let's try to implement/rfc this for a specific usecase and discuss the idea more concretely and if we find multiple places where this idea works we can have a better idea of how it works on a lower level.

masad-frost: Hopefully most og this is all irrelevant once we finish building the system that provides guaranteed eventual consistency

masad-frost:
To stress on what you said the other day, you mentioned we don't really have ACID guarantees anywhere, so "transaction" is the wrong term. I actually think if we call it requestGroup or batch or bundle or something might clear up the conversation

gburtini:
Yeah, I'm good with a rename, though if we can find a way, I'd like to push down this explicitly tagged retry behavior type thing straight to send/request (and as discussed elsewhere, in a later project, right to the protocol level), then the request grouping piece is basically an accessory. We just didn't quickly come up with a way to do that the other day.

src/channel.ts
gburtini: Encoding a `retryArguments` invariant is also an option. It's starting to feel like a bit of a smelly abstraction/design-for-problems-we-don't-have at that point, but it _could_ let us insist on a user actually **making** this decision with the type system.

gburtini: It might be sufficient to just document invariant that way and then make it non-optional.

src/channel.ts
```diff
@@ -2,6 +2,32 @@ import { api } from '@replit/protocol';
...
+  constructor(private error: Error) {
+    super(error.message);
+  }
+}
+
+class InvariantViolation extends Error {
+  constructor(private error: Error) {
+    super(error.message);
+  }
+}
```
gburtini, line 29: This is wrong. Should pass the whole Error subclass through.

gburtini:
Abandon for now, this was too pessimistic on the timeline for alternatives and time spent overcoming the concerns is better spent on getting the alternative out.

PR #186, opened 2023-06-20
[:lady_beetle:] Prevent `openChannel()` from returning an unusable channel.

gburtini:
Why
===

This was mostly an oversight on 10.0 where we had channels never return error states. 

This path can return a closed channel if you called close channel or disconnected while it was opening, this gets us into interlaced connections sometimes for a React-y state reason, but it primarily just causes weird issues in tests.

It's a bit of a hack, but it is better to just _do nothing_ here as it aligns with the intent of the error state refactor. I've pnpm link'd this and had no interesting challenges.

### DIFF
--- a/src/__tests__/index.test.ts
+++ b/src/__tests__/index.test.ts
@@ -778,7 +778,7 @@ concurrent(
 );
 
 concurrent(
-  'allows opening channel with the same name after others are closing others and client is connected (wait for open)',
+  'allows opening channel with the same name after closing (connected, wait for open)',
   (done) => {
     const client = getClient(done);
 
@@ -864,7 +864,7 @@ concurrent(
 );
 
 concurrent(
-  'allows opening channel with the same name after others are closing others and client is connected (do not wait for open)',
+  'allows opening channel with the same name after closing (connected, do not wait for open)',
   (done) => {
     const client = getClient(done);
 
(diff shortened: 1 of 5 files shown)


gburtini:
Gotta get the tests running again before we ship this stuff. Crosis tests are always broken. :-1:

moudy (approved):
🤞

gburtini:
Tests should be fixed here: https://github.com/replit/crosis/pull/188

Rebasing on top of that.

gburtini:
The changes necessary to make the test pass here in Crosis are minor, but important, so I went back and reran the linked version a bit more carefully.

Running linked unit tests on the services directory is promising. The whole suite filled 64GB of RAM and ground to a halt, but I tested a selection of core services and container and they're all passing; in the end I ran just over half the test suites, and everything for fs and container.

src/__tests__/index.test.ts
```diff
@@ -779,7 +778,93 @@ concurrent(
 );
 
 concurrent(
-  'allows opening channel with the same name after others are closing others and client is connected',
+  'allows opening channel with the same name after closing (connected, wait for open)',
+  (done) => {
+    const client = getClient(done);
+
+    console.log('opening client');
```
gburtini, line 785:
```suggestion
```

PR #187, opened 2023-06-20
[docs] Various polish / docs / spelling mistake stuff.

gburtini:
Why
===

I had a long layover in Vancouver. Faris says we're doing docs. This is probably a bit better.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -747,8 +747,6 @@ export class Client<Ctx = null> {
    * - If there's an open WebSocket connection it will be closed
    * - Any open channels will be closed
    *   - Does not clear openChannel requests
-   *   - If a channel never opened, its {@link OpenChannelCb | open channel callback}
-   *     will be called with an error
    *   - Otherwise returned cleanup callback is called
    *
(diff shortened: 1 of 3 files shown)


masad-frost (approved):
haha, nice

src/client.ts
```diff
@@ -1280,12 +1274,12 @@ export class Client<Ctx = null> {
...
-     * the connection if we don't get READY from the container
-     * within the specified timeout.
+     * If the container stops responding for `timeout` period, we will throw.
+     * The container is expected to regularly send status updates if it is still
+     * working and healthy.
      *
-     * Every time we get a message we reset the connection timeout
-     * this is because it signifies that the connection will eventually work.
+     * Every time we get a message we reset the connection timeout this is because
+     * it signifies that the connection should eventually work.
```
masad-frost, line 1282: will->should :grimacing: lol

gburtini: _will_ wasn't correct, especially now that we're blindly sending status update events every 500ms.

gburtini: it pretty much just means "this conman is still alive", not "this connection is still making progress."

PR #188, opened 2023-07-02
[:broom:] Make tests run again.

gburtini:
Why
===

Effectively all tests are failing because they hit Goval without a required token property added here: https://github.com/replit/goval/pull/9461

```
[Symbol(kType)]: 'close',
[Symbol(kCode)]: 1008,
[Symbol(kReason)]: 'missing user id',
[Symbol(kWasClean)]: true
```

What changed
============

Bumped the protocol dependency and added the required property to the fake test metadata.

Test plan
=========

Run the tests.

### DIFF
--- a/debug/genConnectionMetadata.js
+++ b/debug/genConnectionMetadata.js
@@ -30,6 +30,10 @@ function genConnectionMetadata(options) {
     language: 'bash',
     slug: Math.random().toString(36).slice(2),
     user: 'crosistest',
+    userId: {
+      id: 8,
+      environment: 'crosistests',
+    },
     bucket: 'test-replit-repls',
   };
 
--- a/debug/genConnectionMetadata.js
+++ b/debug/genConnectionMetadata.js
@@ -31,7 +31,7 @@ function genConnectionMetadata(options) {
     slug: Math.random().toString(36).slice(2),
     user: 'crosistest',
     userId: {
-      id: 8, // arbitrary
+      id: 78171400, // arbitrary (chosen as crc32c("crosis") if you care)
       environment: api.repl.Environment.DEVELOPMENT,
     },
     bucket: 'test-replit-repls',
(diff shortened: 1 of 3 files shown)


lhchavez (approved):

debug/genConnectionMetadata.js
```diff
@@ -30,6 +30,10 @@ function genConnectionMetadata(options) {
     language: 'bash',
     slug: Math.random().toString(36).slice(2),
     user: 'crosistest',
+    userId: {
+      id: 8,
+      environment: 'crosistests',
```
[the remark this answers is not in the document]
gburtini: Ack.

A random number chosen by human (this is why I tagged you though, I have no idea what this is actually used for). I'll add that as a comment unless you suggest changing it, I just figured 0 and 1 might have special meaning.

debug/genConnectionMetadata.js
```diff
@@ -30,6 +30,10 @@ function genConnectionMetadata(options) {
     language: 'bash',
     slug: Math.random().toString(36).slice(2),
     user: 'crosistest',
+    userId: {
+      id: 8,
```
gburtini, line 34:
```suggestion
      id: 8, // arbitrary
```

lhchavez: if it's arbitrary, can we make it arbitrarier? like the crc32-d hash of `crosis`.

gburtini: Sure.

debug/genConnectionMetadata.js
```diff
@@ -30,6 +30,10 @@ function genConnectionMetadata(options) {
     language: 'bash',
     slug: Math.random().toString(36).slice(2),
     user: 'crosistest',
+    userId: {
+      id: 8, // arbitrary
```
gburtini, line 34:
```suggestion
      id: 78171400, // arbitrary (chosen as crc32c("crosis") if you care)
```

lhchavez (approved):

PR #189, opened 2023-07-02
[:lady_beetle:] Interlaced connections.

gburtini:
Why
===

[Sentry](https://replit-kq.sentry.io/issues/4222002060/?query=is%3Aunresolved+client+entered&referrer=issue-stream&sort=betterPriority&statsPeriod=14d&stream_index=1), among a bunch of other bugs and unhandled exceptions.

~In both cases, I think the `waitingMetadata` was expected to know to return aborted, but I don't understand why both sides of the user API are required to know what is aborted (and I think we're doing it wrong in the wild?).~ 

~(let's move that discussion to #192 unless we need to deal with it here)~ shipped in #193

It is easy enough to get into an interlaced struggle here. The test [here](https://github.com/replit/crosis/pull/189/files#diff-743bff3896313c35c8c7a9d9cb7b8e31f16873dfe58476bfc06dd5c7951de30fR34) shows how. This associates fetch abort controllers with a `connectionId` (just an incrementing number on every open call and explicit disconnect/reconnect) rather than being global to the `Client` instance.

What changed
============

- Add failing tests.
- Make tests pass, well, almost.

Test plan
=========

~Make tests pass... or if we decide we like the two-sided abort thing, at least make the first test throw something other than "client entered wrong state during connect."~

~The state we're in here is best documented by [this test](https://github.com/replit/crosis/pull/189/files#diff-743bff3896313c35c8c7a9d9cb7b8e31f16873dfe58476bfc06dd5c7951de30fR84-R126). It throws right now because the `waitingPromise` has already resolved by the time it is actually aborted.~

~In the wild, both of these requests are actually HTTP requests to the metadata service, so we can't control when they resolve, but they also won't be deferring back to the synchronous thread unless they have multiple promises without checking the abort controller.~

~Not sure what to do about this _in this test_, but my thought is coming back around to removing the assertion entirely.~

All of the above was fixed by #193 which is shipped all the way to web at this point.

Tests written, tests passed. I have _not_ yet wired this to web, will do before requesting review; the main test there will be ensuring the ReplViewer interlaced connection doesn't repro.

### DIFF
--- a/src/__tests__/regressions.test.ts
+++ b/src/__tests__/regressions.test.ts
@@ -67,7 +67,7 @@ concurrent("interlaced connections don't throw", async (done) => {
   client.open(
     { ...params, fetchConnectionMetadata: abortingMetadata },
     wrapWithDone(done, () => {
-      console.log('first open');
+      throw new Error('Expected to never get here (abort+instant close).');
     }),
   );
   // set up the initial abort
@@ -76,7 +76,7 @@ concurrent("interlaced connections don't throw", async (done) => {
   client.open(
     { ...params, fetchConnectionMetadata: waitingMetadata },
     wrapWithDone(done, () => {
-      console.log('second open');
+      console.log('reached second client.open');
     }),
   );
 
--- a/src/__tests__/regressions.test.ts
+++ b/src/__tests__/regressions.test.ts
@@ -1,0 +1,86 @@
+/* eslint-env jest */
+
+import { Client, FetchConnectionMetadataError, FetchConnectionMetadataResult } from '..';
+import { concurrent } from '../__testutils__/concurrent';
+import { wrapWithDone } from '../__testutils__/done';
+import genConnectionMetadata from '../../debug/genConnectionMetadata';
+
+// eslint-disable-next-line
+const WebSocket = require('ws');
+
+jest.setTimeout(30 * 1000);
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const testingClients: Array<Client<any>> = [];
+
+// Just a helper that to help us exit from jest without any open handles
+function getClient<Ctx = null>(done: jest.DoneCallback) {
+  const c = new Client<Ctx>();
+  c.setUnrecoverableErrorHandler(done);
+  testingClients.push(c);
+
+  return c;
+}
+
+const getConnectionMetadata = async () => ({
+  ...genConnectionMetadata(),
+  error: null,
+});
+
+afterAll(() => {
+  testingClients.forEach((c) => c.destroy());
+});
+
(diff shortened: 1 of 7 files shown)


src/__tests__/regressions.test.ts
```diff
@@ -0,0 +1,166 @@
...
+  const waitingMetadata = () => {
+    if (waitingPromise) {
+      return waitingPromise;
+    }
+
+    waitingPromise = new Promise<FetchConnectionMetadataResult>((resolve) => {
+      return getConnectionMetadata().then(resolve);
+    });
+    return waitingPromise;
+  };
```
gburtini, line 120: This one is less clear that it is an actual 'bug' (vs. just missed encapsulation) because if you replace it with the 'correct' implementation, everything works as expected. The error message is also clear that's the case.

```suggestion
  let waitingPromise: Promise<FetchConnectionMetadataResult> | null = null;
  const waitingMetadata = (abortSignal: AbortSignal) => {
    if (waitingPromise) {
      return waitingPromise;
    }

    waitingPromise = new Promise<FetchConnectionMetadataResult>((resolve) => {
      return getConnectionMetadata().then((r) => {
        if (abortSignal.aborted) {
          resolve({
            error: FetchConnectionMetadataError.Aborted,
          });
        }

        resolve(r);
      });
    });
    return waitingPromise;
  };
```

gburtini: Going to drop this one, per the point here: https://github.com/replit/crosis/pull/192

gburtini:
Dependent on #193 for now. I can revise the tests accordingly, but I'd like to ship 193 separately from this more intimidating change.

jakehamilton (approved):
Thank you for jumping on this!

src/client.ts
```diff
@@ -1816,7 +1840,12 @@ export class Client<Ctx = null> {
       message: 'status:reconnecting',
     });
 
-    this.connect({ tryCount: 0, websocketFailureCount: 0 });
+    this.connectionId++; // TODO: should will reconnects increment this or not?
```
gburtini, line 1843: should will!?

jakehamilton: does is have will???

gburtini: I can't ship till I answer the question I asked there. On it.

gburtini: oh it isn't as incoherent as I thought either. That's "should `willReconnect`s increment this "

gburtini: At this point, we've unambiguously cleaned up the handler, so incrementing is correct.

```suggestion
    this.connectionId++;
```

bradymadden97 (approved):
Nice test

jackyzha0 (approved):

PR #192, opened 2023-07-05
[:bulb:] Don't require the user to return Aborted when that is just fed to it by Crosis.

gburtini:
Why
===

We largely agreed not to do this, or to do it by removing `FetchConnectionMetadataError.Aborted` entirely, but in case we want the discussion, I'm going to open and immediately close this.

These two signals currently have to match (user and API), which means the only way a client is allowed to abort is if the `abortController` said so. With that in mind, it doesn't make a lot of sense to me to require the client to return aborted if we're going to bail anyway.

This is a change that implements that thought. To ship, I'd rather just delete the concept of the client returning aborted entirely and just have its return value as meaningless if we're aborted.

What changed
============

Test plan
=========

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -1114,34 +1114,11 @@ export class Client<Ctx = null> {
       this.fetchTokenAbortControllers.delete(connectionId);
 
       const connectionMetadata = connectionMetadataFetchResult;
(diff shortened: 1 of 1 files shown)


gburtini:
Superseded by #193.

PR #193, opened 2023-07-06
Tweak the consumer + Crosis both sides aborted check to allow Crosis to unilaterally abort.

gburtini:
Why
===

We discussed _removing_ this check, because it was requiring the client to behave in a particular way w.r.t. receiving an abort signal. This tweak changes that requirement without breaking any existing client usages, keeping an invariant assertion, but making it optional on the _did abort_ case.

What changed
============

Before:
- Crosis aborts.
- Consumer _must_ return Aborted. _It turns out this requirement was impossible to satisfy in certain synchronous contexts._

After:
- Crosis aborts.
- Consumer _may_ return Aborted.

Before:
- Crosis _does not_ abort.
- Consumer _may not_ return Aborted.

After:
- Crosis _does not_ abort.
- Consumer _may not_ return Aborted.

Test plan
=========

I'll wire to web and update the appropriate unit tests, but most of the testing is going to take place by releasing this.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -10,6 +10,12 @@ export enum ConnectionState {
 
 export enum FetchConnectionMetadataError {
   /**
+   * Fetch was aborted. A client can return this if they decided not to fetch
+   * connection metadata _because_ the abort controller was already triggered.
+   */
+  Aborted = 'Aborted',
+
+  /**
    * The fetch failed due to a recoverable error (mostly a transient network
    * condition).
    */
--- a/src/types.ts
+++ b/src/types.ts
@@ -10,11 +10,6 @@ export enum ConnectionState {
 
 export enum FetchConnectionMetadataError {
   /**
-   * Fetch was aborted.
-   */
-  Aborted = 'Aborted',
-
-  /**
    * The fetch failed due to a recoverable error (mostly a transient network
    * condition).
    */
--- a/src/types.ts
+++ b/src/types.ts
@@ -10,8 +10,8 @@ export enum ConnectionState {
 
 export enum FetchConnectionMetadataError {
   /**
-   * Fetch was aborted. A client can return this if they decided not to fetch
-   * connection metadata _because_ the abort controller was already triggered.
+   * Fetch was aborted. A client may return this only if 
+   * the `AbortSignal` was already aborted.
    */
   Aborted = 'Aborted',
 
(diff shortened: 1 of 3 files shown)


src/types.ts
```diff
@@ -10,7 +10,8 @@ export enum ConnectionState {
 
 export enum FetchConnectionMetadataError {
   /**
-   * Fetch was aborted.
+   * Fetch was aborted. A client can return this if they decided not to fetch
+   * connection metadata _because_ the abort controller was already triggered.
```
masad-frost, line 14:
```suggestion
   * Fetch was aborted. A client must return this if the `AbortSignal` passed is aborted.
   * If the client chooses to abort the request, it must call `signal.abort()` and return this.
```

Is

> If the client chooses to abort, it must call `signal.abort()` and return this.

correct? or does it need to be `client.close()`?

masad-frost (approved):
Seems reasonable.

src/types.ts
gburtini: Client calls `client.close`... which I have no idea if you can actually safely do inside a fetchConnectionMetadata call.

[AbortSignal instances](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal) don't have an abort method (the controller has it). I _think_ that confusion is what led to this assertion seeming reasonable in the first place: only Crosis can trigger an abort.

gburtini: The rest of your suggestion feels like it is orthogonal to this change. 

The client _may_ return Aborted now iff we're aborted; it may also return valid metadata that just won't be used. It may _not_ return Aborted if we're not aborted.

This is basically just a convenience handler for valid types.

gburtini:
```suggestion
   * Fetch was aborted. A client may return this only if they decided not to fetch
   * connection metadata _because_ the `AbortSignal` was already aborted.
```

gburtini: Maybe this is better.

masad-frost: Yeah that seems more sane to me.

src/types.ts
```diff
@@ -10,7 +10,8 @@ export enum ConnectionState {
 
 export enum FetchConnectionMetadataError {
   /**
-   * Fetch was aborted.
+   * Fetch was aborted. A client may return this only if 
+   * the `AbortSignal` was already aborted.
```
masad-frost, line 14: I would also add that the it *must* return this if the passed signal was aborted. According to the code at least, it seems that way.

masad-frost: nevermind lol

PR #199, opened 2023-10-17
fix(security): bump babel traverse to avoid cve

jakehamilton:
Why
===

_Describe what prompted you to make this change, link relevant resources: Asana tasks, Canny report, Slack discussions...etc_

https://github.com/replit/crosis/security/dependabot/76

What changed
============

Bumped `@babel/traverse`.

Test plan
=========

This change only affects `jest`. As long as tests pass we're good to go.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,8 @@
     "ts-jest": "^28.0.2",
     "typedoc": "^0.22.15",
     "typescript": "^4.6.4",
-    "ws": "^8.6.0"
+    "ws": "^8.6.0",
+    "@babel/traverse": "^7.23.2"
   },
   "peerDependencies": {
     "@replit/protocol": ">=0.2.15"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -44,7 +44,8 @@
     "ts-jest": "^28.0.2",
     "typedoc": "^0.22.15",
     "typescript": "^4.6.4",
-    "ws": "^8.6.0"
+    "ws": "^8.6.0",
+    "@babel/traverse": "^7.23.2"
```
sergeichestakov, line 48: why is this out of order? when I added this locally it put it at the front behind `@replit/protocol` since this is alphabetically sorted. can we match that here?

sergeichestakov (approved):
sounds reasonable. thanks

package.json
jakehamilton: Ha, yep will do.

PR #202, opened 2023-11-26
chore: add nix flake and envrc

jakehamilton:
Why
===

I wanted to work on Crosis, but we didn't have a dev shell defined like we do in most other projects.

What changed
============

I've added a Nix Flake that provides a dev shell with `yarn`, `python3`, and `nodejs` so that working with the project can be done without needing to manage system dependencies manually. A `.envrc` was also included which enables `direnv` to automatically activate the dev shell when entering the directory.

Test plan
=========

This change does not affect the library in any way. What I did to confirm that this worked was to enter the dev shell with `nix develop` (or direnv) and then attempt to use `yarn` to install dependencies and build the library.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,1 @@
+use flake
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ yarn-error.log*
 .cache/
 
 .env
+.direnv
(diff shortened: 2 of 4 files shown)


ryantm (approved):
Looks good, thanks!

PR #204, opened 2024-03-22
Make destroy always clean up resources

masad-frost:
Why
===

There are cases where we destroy call destroy while disconnected and it leads us to not go through all the cleanup. This change makes it so that we unconditionally call `handleClose` to clean up everything, forever.

What changed
============

Made `destroy` call `handleClose`

Test plan
=========

Tests for destroy pass.


jackyzha0 (approved):

jackyzha0:
image

PR #208, opened 2024-04-11
[dev] Update instructions to run tests

lhchavez:
Why
===

We have for very long relied on folks relying on some form of production infra to test this.

What changed
============

This change now allows folks to test this out with a local checkout of goval.

Test plan
=========

Ran the new instructions, most tests passed!

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -2,8 +2,8 @@ version: v1.0
 name: Node JS
 agent:
   machine:
-    type: e1-standard-2
-    os_image: ubuntu1804
+    type: e2-standard-2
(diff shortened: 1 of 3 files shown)


sergeichestakov (approved):
thank you! should we still support running against prod though for those who don't have local goval setup or those in the OSS community?

lhchavez:
> should we still support running against prod [...]

i would _very much_ not want to allow that D: nobody should be running any kind of tests against production. we'll probably have some form of shared dev environment soon, we can discuss if it's possible to extend that affordance to non-employees (still very yikes, but not _as_ yikes).

in any case https://github.com/replit/river is the new hotness

PR #223, opened 2026-07-01
Surface terminal close codes for taken-down and deleted Repls

jasondellaluce:
Why
===

When a Repl is blocked from running — for example after a policy/abuse takedown — or has been deleted, the infrastructure closes the websocket with a dedicated close code (`4011` for a takedown, `4010` for a deletion). The client only recognized a handful of close codes, so these fell through to the default branch and were treated as a transient, retriable failure. As a result the client would silently reconnect against the same block until the connection timed out, and the UI never learned that the Repl can't run so it couldn't explain why to the user.

What changed
============

- Added `REPL_DELETED` (`4010`) and `REPL_TAKEN_DOWN` (`4011`) to the `CloseCode` map in `src/types.ts`.
- Handled both codes in the websocket `onclose` path (`src/client.ts`): they are now non-retriable and produce a descriptive `errorMessage` (Terms-of-Service takedown / deleted Repl) instead of the generic `WebSocket closed before we got READY`, mirroring how `USER_ERROR`, `FIREWALL_DENIED`, and `CONCURRENT_REPL_LIMIT` are already handled.
- Forwarded the websocket `closeCode` (and `reason`, when present) as `extras` on the emitted `CrosisError`, so consumers can distinguish the underlying cause and log it.
- Added a retry test asserting a `4011` close does not retry and surfaces a takedown message.

This is additive (two new close codes + handling) and backwards compatible; unrecognized codes still behave as before.

Test plan
=========

- `yarn lint` and `yarn build` pass.
- Added `should not retry when the Repl is taken down` to `src/__tests__/retry.test.ts`, mirroring the existing `should not retry for user error` case: closing with `REPL_TAKEN_DOWN` yields a single connection attempt, a non-retriable unrecoverable error, and a message containing "taken down".

Note: the `src/__tests__/retry.test.ts` suite does not run cleanly in isolation locally (the pre-existing cases time out / collide on the mock websocket port under the current `jest-websocket-mock`/`mock-socket` setup), so I could not get a green local run of that file. The new case follows the exact pattern of the existing `user error` test.


jasondellaluce:
Closing this — on reflection this is the wrong layer to surface a repl takedown.

The abuse/tombstone signal is a property of the repl, and both the PROTOBUF (crosis) and RIVER (pid2) clients connect through the same `/wsv2` endpoint and receive the same close codes. Handling it here would only cover the crosis transport and miss the River path, and both transports already converge on a shared connection-state layer downstream. We're going to detect the takedown once at the transport-agnostic connection-metadata step instead, so both clients get consistent handling without per-transport close-code logic.

The test-helper fix in here (setting `wsURL` so the mock-socket suites hit the mock server after #221) is still valid and worth landing separately if useful.

